Issue #29, opened 2023-03-10
Inject `GOBIN` into the user's shell 


milesj:
If a users starts using go via proto without running the official go installer, the `GOBIN` env var will not be set. We should inject this automatically into the user's shell.

PR #38, opened 2023-03-14
new: Set `GOBIN` in shell profile when installing Go for the first time.

milesj:
If a user installs Go through proto for the first time, and Go has not been installed via homebrew or the official installer, the `GOBIN` env var is not set, resulting in Go binaries not being located. This PR adds support for setting `GOBIN` to the user's shell.

This uses the same code from the `setup` command, so I reworked the shell logic to generic.

Implements https://github.com/moonrepo/proto/issues/29


crates/cli/src/hooks/go.rs
```diff
@@ -0,0 +1,32 @@
...
+        return Ok(());
+    }
+
+    let Some(shell) = Shell::from_env() else {
+        return Ok(());
+    };
+
+    let env_vars = FxHashMap::from_iter([
+        ("GOBIN".to_string(), "$HOME/go/bin".to_string()),
+        ("PATH".to_string(), "$GOBIN".to_string()),
```
jpoz, line 20: Thought his was setting the PATH to GOBIN. A little confusing that's a _special_ key, but I get it

jpoz (approved):

crates/cli/src/hooks/go.rs
milesj: I feel that. Felt the same while writing it but this was the simplest approach.

PR #68, opened 2023-04-09
Allow installing multiple global binaries

### LINKED ISSUE #66: Allow installing multiple global binaries
It is very tedious to write:

```bash
proto install-global node wrangler && proto install-global node typescript-language-server && proto install-global node vscode-langservers-extracted
```

Would be really nice if that could be condensed to:

```bash
proto install-global node wrangler typescript-language-server vscode-langservers-extracted
```

IgnisDa:
Fixes #66

### DIFF
--- a/crates/cli/src/commands/install_global.rs
+++ b/crates/cli/src/commands/install_global.rs
@@ -13,108 +13,113 @@ async fn get_bin_or_fallback(mut tool: Box<dyn Tool<'_>>) -> Result<PathBuf, Pro
     })
 }
 
-pub async fn install_global(tool_type: ToolType, dependency: String) -> Result<(), ProtoError> {
+pub async fn install_global(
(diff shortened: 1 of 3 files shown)


crates/cli/src/commands/install_global.rs
```diff
@@ -13,108 +13,113 @@ async fn get_bin_or_fallback(mut tool: Box<dyn Tool<'_>>) -> Result<PathBuf, Pro
...
-                    "install",
-                    "--global",
-                    "--loglevel",
-                    "warn",
-                    "--no-audit",
-                    "--no-update-notifier",
-                ])
-                .arg(&dependency)
-                .env("PREFIX", &global_dir);
+    for dependency in dependencies {
```
milesj, line 22: its unfortunate that these commands dont support multiple, but this is fine

Issue #85, opened 2023-04-24
`Failed to parse JSON file ~/.proto/tools/node/manifest.json` when running multiple tasks


konomae:
Occasionally the following errors may occur. 🤔
maybe, related: https://github.com/moonrepo/proto/pull/75

---

My package.json:

```json
{
  "name": "proto-problem",
  "version": "1.0.0",
  "scripts": {
    "task": "run-p task:*",
    "task:a": "node -e \"console.log('a')\"",
    "task:b": "node -e \"console.log('b')\"",
    "task:c": "node -e \"console.log('c')\"",
    "task:d": "node -e \"console.log('d')\"",
    "task:e": "node -e \"console.log('e')\"",
    "task:f": "node -e \"console.log('f')\"",
    "task:g": "node -e \"console.log('g')\"",
    "task:h": "node -e \"console.log('h')\"",
    "task:i": "node -e \"console.log('i')\"",
    "task:j": "node -e \"console.log('j')\""
  },
  "devDependencies": {
    "npm-run-all": "^4.1.5"
  }
}
```

and proto version 0.7.0:

```shell
proto --version
proto 0.7.0
```

then, run the task:

```shell
npm install
npm run task
```

and the result:

```
Error: proto::json
  × Failed to parse JSON file ~/.proto/tools/node/manifest.json
  ╰─▶ EOF while parsing a value at line 1 column 0
```

milesj:
Can you try v0.7.1?

konomae:
@milesj 
It was fixed! 🎉
Thank you!

milesj:
Awesome

TheUeyOfTheForce:
Getting the same error:

```
Error: json::parse_file

  × Failed to parse JSON file ~/.proto/tools/node/manifest.json.
  ╰─▶ EOF while parsing a value at line 2 column 0```

Proto version: `0.15.1`

**Comment by YishaqG** [comment]
Getting the same error:
```
Error: json::parse_file

  × Failed to parse JSON file ~/.proto/tools/node/manifest.json.
  ╰─▶ EOF while parsing a value at line 2 column 0
```
Proto version: `0.15.1`

milesj:
@YishaqG What commands are you running to hit it?

YishaqG:
@milesj Just running `node

YishaqG:
@milesj Just running `node` or `npm`, was working with `worker_threads` module and after a couple of runs I got that error.

milesj:
[...]

PR #87, opened 2023-04-24
fix: Improve file locking logic.

milesj:
Fixes in comments.

Fixes https://github.com/moonrepo/proto/issues/85

### DIFF
--- a/crates/core/src/manifest.rs
+++ b/crates/core/src/manifest.rs
@@ -101,23 +101,7 @@ impl Manifest {
         debug!(file = %path.display(), "Loading manifest");
 
         let mut manifest: Manifest = if path.exists() {
-            use fs4::FileExt;
-
-            let file = fs::open_file(path)?;
-
-            file.lock_shared().map_err(|error| FsError::Read {
-                path: path.to_path_buf(),
-                error,
-            })?;
-
-            let data = json::read_file(path)?;
-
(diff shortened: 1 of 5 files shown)


crates/core/src/manifest.rs
```diff
@@ -130,32 +135,40 @@ impl Manifest {
...
-        use std::io::prelude::*;
+        use std::io::{prelude::*, SeekFrom};
 
         debug!(file = %self.path.display(), "Saving manifest");
 
         if let Some(parent) = self.path.parent() {
             fs::create_dir_all(parent)?;
         }
 
-        let mut file = fs::create_file(&self.path)?;
```
milesj, line 141: This would truncate the file, and then other processes would read an empty file causing failures.

crates/core/src/manifest.rs
```diff
@@ -130,32 +135,40 @@ impl Manifest {
...
+        file.lock_exclusive().map_err(handle_error)?;
 
         let data = json::to_string_pretty(&self).map_err(|error| JsonError::StringifyFile {
             path: self.path.to_path_buf(),
             error,
         })?;
 
-        write!(file, "{}", data).unwrap();
+        // Truncate then write file
+        file.set_len(0).map_err(handle_error)?;
```
milesj, line 166: Only truncate after we have acquired the lock

PR #103, opened 2023-05-17
new: Add experimental WASM plugin support.

milesj:


crates/wasm-plugin/src/resolve.rs
```diff
@@ -0,0 +1,24 @@
...
+impl Resolvable<'_> for WasmPlugin {
+    fn get_resolved_version(&self) -> &str {
+        match self.version.as_ref() {
+            Some(version) => version,
+            None => "latest",
+        }
+    }
+
+    async fn load_version_manifest(&self) -> Result<VersionManifest, ProtoError> {
+        // let result = self.call_method("hello_world", "{}");
```
jpoz, line 14: Hi 👋

crates/wasm-test/src/lib.rs
```diff
@@ -0,0 +1,60 @@
+use extism_pdk::*;
+use proto_plugin_sdk::*;
+
+#[plugin_fn]
+pub fn get_metadata(_: ()) -> FnResult<Json<PluginMetadata>> {
```
jpoz, line 5: So as long as the plugin functions are named like this, it will work?

Do we check if these functions are around when we load the wasm plugin?

crates/wasm-plugin/tests/wasm_plugin_test.rs
```diff
@@ -0,0 +1,91 @@
...
+    };
+
+    let mut tool = WasmPlugin::new(
+        Proto::from(dir),
+        "wasm".into(),
+        wasm_dir.join("proto_wasm_test.wasm"),
+    )
+    .unwrap();
+
+    tool.version = Some("1.0.0".into());
```
jpoz, line 29: Why do we need to set a version for a plugin?

Issue #122, opened 2023-06-05
can't install node 14 on m1 macs


hayes:
Node does not have a `darwin-arm64` build for node 14, which causes installation to fail on m1 macs

milesj:
@hayes Do you not receive an error or anything? What exactly is happening?

hayes:
<img width="1092" alt="Screenshot 2023-06-05 at 10 32 11 AM" src="https://github.com/moonrepo/proto/assets/816527/739f4c8a-1599-4bfc-bda5-57ffc2a91d0f">

Looks like https://nodejs.org/dist/v14.21.3/node-v14.21.3-darwin-arm64.tar.xz is looking for the native m1 build, but I don't think node starting creating builds for m1 until node 16 (https://nodejs.org/dist/v14.21.3/)

milesj:
Ok that's the error I expected. I can improve the message a bit for 404s.

hayes:
@milesj does that mean you can't use proto if you want to use node 14?  Seems like it should just install the intel build for node versions before 16 right?

hayes:
yup, it works great.  It's slightly slower because it will run through rosetta, but basically anything build for intel will run fine

PR #126, opened 2023-06-05
Fix installs for node@<16 on m1 macs

### LINKED ISSUE #125: install should add a newline after adding executable to PATH
proto adds the following to your profile/zshrc

```
# proto
export PROTO_ROOT="$HOME/.proto"
export PATH="$PROTO_ROOT/bin:$PATH"
```

but it doesn't add a trailing newline.  The can cause issues when other tools don't add prepend a newline before they make their own changes.

hayes:
fixes #125 and #122 

Screenshot 2023-06-05 at 2 53 56 PM


milesj:
@hayes Awesome, was easier than expected.

Issue #120, opened 2023-06-05
New Language: python


bjesuiter:
Hi there, 

one language, that would be extremely nice to have in proton is python! 
Especially seeing the problems around python 2 and python 3. 
Maybe you can integrate pyenv internally or something like this, like with rust. 
Or you decide, that it would be worth it to have direct python support, 
even though I'm imagining that this might be hard to do.

milesj:
Landed in 0.17!

PR #141, opened 2023-07-03
new: Add Python language support

ericmccarthy7:
Implements #120 

---

Hello!

Took a stab at implementing python support in proto. I would like to use python in my personal projects using moon. This hopefully brings me closer to that goal (I will work on adding to moon if this PR is eventually accepted)

This is pretty opinionated since I am using [rye](https://github.com/mitsuhiko/rye/). I figured it would save me a bunch of time from trying to reimplement all of the features already implemented there.

I know this PR needs a ton of work so I opened as a draft. I am new to rust and moonrepo so I would love any feedback that can be provided!

Before this should be seriously considered I will:

- [x] add available versions instead of a single hardcoded one (quickly used to understand architecture)
- [x] get it working on multiple platforms (only tested on mac so far)
- [ ] add tests (any feedback on what I should test? if none I'll base on `crates/rust/tests`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1693,6 +1693,7 @@ dependencies = [
  "proto_deno",
  "proto_go",
  "proto_node",
+ "proto_python",
  "proto_rust",
  "proto_schema_plugin",
  "reqwest",
@@ -1774,6 +1775,20 @@ dependencies = [
  "proto_core",
  "rustc-hash",
  "serde",
+ "starbase_sandbox",
+ "starbase_utils",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
+name = "proto_python"
+version = "0.1.0"
+dependencies = [
+ "once_cell",
+ "proto_core",
+ "serde",
+ "serde_json",
  "starbase_sandbox",
  "starbase_utils",
  "tokio",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -32,6 +32,7 @@ proto_core = { version = "0.11.1", path = "../core" }
 proto_deno = { version = "0.10.0", path = "../deno" }
 proto_go = { version = "0.10.0", path = "../go" }
 proto_node = { version = "0.10.0", path = "../node" }
+proto_python = { version = "0.1.0", path = "../python" }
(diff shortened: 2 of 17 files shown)


milesj:
@ericmccarthy7 Thanks for this initial spike. I rarely use Python, so haven't had the time to invest into this yet.

I'm not entirely sure how I feel about using another version manager to power this version manager. I know this is based on the Rust implementation, but that's a bit of a special case as there's no other way to use Rust, but that's not the case for Python.

With that said, I was looking at Rye's source code, and it looks like it simply downloads pre-builts from https://github.com/indygreg/python-build-standalone/releases I feel like proto can do the same thing here.

ericmccarthy7:
@milesj I understand why you would not want to rely on another version manager - that being said I would like to still go through the learning exercise of implementing using rye. I think the real benefit is when you get to package/dependency management.

I am apprehensive to try this with just the raw binaries because of the confusing nature of virtual environments, global dependencies, `pyproject.toml` vs `requirements.txt`, etc. We would be making some opinionated choices anyways and I feel that the ones made in `rye` line up with `proto`/`moon`. 

If you feel this is a wasted effort maybe I should rethink this as a plugin when you release that capability? I see in the docs moon will eventually support language plugins but not sure how to approach it (https://moonrepo.dev/docs/how-it-works/languages)

ericmccarthy7:
This actually appears to be working just fine. Only addition left is tests.

Only thing that's not super nice is that we use the local version of rye to get latest available python versions. Might be worth calling `rye self update` first to make sure rye is up to date.

milesj:
@ericmccarthy7 Yeah, maybe waiting on this may be best. WASM plugins will land shortly, maybe in the next week or 2, and this rye approach sounds much more suited as a WASM plugin instead of built-in core.

This is the WASM progress: https://github.com/moonrepo/proto/pull/103 All that's left is version resolution.

milesj:
@ericmccarthy7 WASM support is out: https://moonrepo.dev/blog/proto-v0.12

milesj:
Started working on python support for v0.17 here: https://github.com/moonrepo/python-plugin

This doesn't use rye and instead uses the pre-builts. If you really want to use rye, I suggest creating your own plugin.

PR #156, opened 2023-07-21
fix: Remove space after gt lt gte lte in detector

mmiszy:
Replace whitespace after `>`, `>=`, `<` and `<=` in `detector`.

It's how `node-semver` handles it: https://github.com/npm/node-semver/blob/cce61804ba6f997225a1267135c06676fe0524d2/internal/re.js#L188-L192


milesj:
@mmiszy Awesome stuff, looks good. Just need to fix the formatting.

PR #272, opened 2023-11-06
v0.22

milesj:


rotu:
I'm a little iffy on the concept of a parent bin. If I'm running a script with `npm run`, and that path contains `node myscript.js`, we probably need that `node` invocation to resolve to the same .exe as whatever magic determines the parent executable.

milesj:
@rotu Do you have an actual example?

I tried using the shell scripts provided by npm/yarn, like `bin/npm` instead of `bin/npm-cli.js`, but they were completely broken as those scripts do a lot of weird path detection stuff, which isn't compatible with proto. So I have to execute the JS files used the parent (node).

PR #305, opened 2023-11-23
fix: Fix listing globals. Update plugins.

milesj:
@rotu Fixes things.

One caveat though is that Python bins and globals are the same dir, so listing globals lists a bunch of other stuff. Not sure we can work around that easily.

Screenshot 2023-11-22 at 4 22 40 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@
 - [Rust](https://github.com/moonrepo/rust-plugin/blob/master/CHANGELOG.md)
 - [TOML schema](https://github.com/moonrepo/schema-plugin/blob/master/CHANGELOG.md)
 
+## Unreleased
+
+#### 🐞 Fixes
+
+- Fixed `proto list-global` not resolving a version.
+
 ## 0.23.3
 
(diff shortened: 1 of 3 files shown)


rotu:
> One caveat though is that Python bins and globals are the same dir, so listing globals lists a bunch of other stuff.

A “global” could mean the name of a package, a module, or a binary. What do you intend `proto` to mean by that word for Python?

milesj:
`list-global` is meant to display global packages that the user has ideally installed with `install-global`. Showing more than what they installed may confuse users.

rotu:
> `list-global` is meant to display global packages that the user has ideally installed with `install-global`. Showing more than what they installed may confuse users.

Hmmm that’s a hard problem! `pip` doesn’t track which dependencies are directly requested. And listing out some bin folder won’t do since a package may have zero, one, or several executables.

I’ve enjoyed using `poetry` in the past to specify projects or environments. I’ve found pipenv to be a terrible experience (but I don’t remember why). And a `requirements.txt` file is probably the lightest-weight way to do it (with something like `pip-sync`)

Issue #311, opened 2023-11-30
proto run npm/pnpm/yarn fails with `program not found` if bin not in path


rotu:
### What version?

0.23.5

### What happened?

```console 
> proto run pnpm
Error:   × program not found
```

My path includes `~/.proto/shims` but *NOT* `~/.proto/bin` per [here](https://github.com/moonrepo/proto/issues/269#issuecomment-1832924875).

node_depman_plugin version 0.5.3
node_plugin version 0.5.3

### Trace logs?

```console
$ proto run pnpm
[DEBUG 2023-11-30 11:23:44] proto  Running proto v0.23.5  args=["C:\\Users\\dan\\.proto\\shims\\proto.exe", "run", "pnpm", "--log", "trace"]
[TRACE 23:44.785] starbase::app  Running startup phase with 1 systems
[TRACE 23:44.785] starbase::app  Running analyze phase with 1 systems
[TRACE 23:44.786] starbase::app  Running execute phase with 2 systems
[DEBUG 23:44.790] proto_core::user_config:run  Loading config.toml  file="C:\\Users\\dan\\.proto\\config.toml"
[TRACE 23:44.791] starbase_utils::fs:run  Reading file  file="C:\\Users\\dan\\.proto\\config.toml"
[DEBUG 23:44.791] proto_core::tool_loader:run  Traversing upwards to find a configured plugin  tool="pnpm"
[DEBUG 23:44.792] proto_core::tools_config:run  Loading .prototools  file="C:\\Users\\dan\\.prototools"
[TRACE 23:44.792] starbase_utils::fs:run  Reading file  file="C:\\Users\\dan\\.prototools"
[TRACE 23:44.802] warpgate::loader:run  Creating plugin loader  cache_dir="C:\\Users\\dan\\.proto\\plugins"
[TRACE 23:44.802] warpgate::loader:run  Loading plugin pnpm  plugin="pnpm"
[TRACE 23:44.803] warpgate::loader:run  Plugin already downloaded and cached  plugin="pnpm" path="C:\\Users\\dan\\.proto\\plugins\\pnpm-50fe3b6c66bfa9dd0663411862e9dd8fb7e301f8387f8dcd03a890094c977902.wasm"
[...]

PR #315, opened 2023-12-01
new: Create a new shim executable.

milesj:


rotu:
Go Miles! I'm super excited to have a shim solution that's not specific to the shell!

rotu:
> @rotu So far this works perfectly for primary shims (npm, node, etc), but unfortunately doesn't work for secondary shims (npx, pip, etc).
> 
> The problem is that secondary shims add `--alt bin` flags and before/after args into the shell file, which are then passed to `proto`. When relying on `argv[0]`, we have lost this information.
> 
> The only solution I can think of is having some kind of manifest like `~/.proto/shims/manifest.json` that keeps a mapping of how everything should be executed.

Agreed, but I think secondary shims are a problem from the beginning. That's what I meant [in my comment](https://github.com/moonrepo/proto/issues/311#issuecomment-1834632142):

> 3. To that end, `proto` needs a way to map the name of a bin to the plugin/tool which provides it. I realize this is ambiguous (e.g. two plugins could declare a program with the same name) and `proto` needs to handle this either implicitly or by saving desired shims in the `.prototools` config.

I think the easiest thing is to add a section to the yaml, which maps the name to a `$PROTO_HOME`-relative path:

```yaml
[shims]
npm = "tools/npm/10.2.4/bin/npm"
```

This even provides the ability to shim to the *real* bundled `npm` instead of one provided by `node_depman_plugin`:

```yaml
[shims]
npm = "tools/node/20.10.0/bin/npm"
```

If these need to float with the version, the path could instead be, e.g. `tools/npm/${PROTO_NPM_VERSION}/bin/npm` or use some other placeholder like `proto:npm/bin/npm`.

milesj:
Yup exactly. Will look into ways to easily solve this. This may land in 0.25 instead of 0.24. Want to ensure it all works correctly.

rotu:
This also allows you to e.g. remap `node` like `bunx --bun` does:
```
[shims]
node = path/to/bun
```
or even possibly configure arguments on a per-project basis:
```
[shims]
node = ["path/to/node", "--loader", "ts-node/esm"]
```

PR #329, opened 2023-12-13
v0.26

milesj:


rotu:
I'm rather excited for this! Finally going to be able to ditch my PowerShell profile hacks!

milesj:
@rotu I think this is ready to start testing. Would appreciate some help testing this before releasing it officially.

You can download the new bins from the canary tag https://github.com/moonrepo/proto/releases/tag/canary (once this job is done https://github.com/moonrepo/proto/actions/runs/7216349848).

Simply download the bins, rename them to `proto.exe` and `proto-shim.exe`, put them in `~/.proto/bin` (or `PROTO_INSTALL_DIR`), and give it a whirl.

rotu:
I'm running this on macOS right now. So far it seems to work remarkably well!

I did find VSCode has some trouble - notably if I try to open a "JavaScript Debug Terminal", it doesn't seem to work. Haven't gotten to a root cause yet.

In trying to debug this, I found that the shims *do* require `proto` to be in the `PATH`. If not, you get this unhelpful message:

```ShellSession
$ env -i PROTO_LOG=trace /Users/dan/.proto/shims/node --version
[TRACE 2023-12-14 22:56:39] proto_shim  Running node shim  shim="node" args=["/Users/dan/.proto/shims/node", "--version"] file="/Users/dan/.proto/shims/node"
[TRACE 56:39.638] proto_shim  Spawning proto child process  shim="node"
Error: No such file or directory (os error 2)
```

milesj:
I'll add some more custom error messaging for situations like this.

rotu:
> I did find VSCode has some trouble - notably if I try to open a "JavaScript Debug Terminal", it doesn't seem to work. Haven't gotten to a root cause yet.

Okay did a little more digging on this and it also seems like running a JavaScript file from the debug menu causes VSCode to think the process is still launching. This is the case when my settings has `"debug.javascript.defaultRuntimeExecutable": { "pwa-node": "/Users/dan/.proto/shims/node" }` but NOT when my settings bypass the shim with `"debug.javascript.defaultRuntimeExecutable": { "pwa-node": "/Users/dan/.proto/tools/node/20.10.0/bin/node" }`.

rotu:
And it seems to have an issue as well detecting the Python version (3.12 in my case).
image

I'm thinking this could be due to the same reliance on `proto`. It does not seem unreasonable to simply ignore `$PATH` and expect the `proto` binary to be colocated with `proto-shim`.

milesj:
@rotu Trying here https://github.com/moonrepo/proto/pull/341

rotu:
(fwiw, I think it's easier to have this be available on the PR action rather than as a release with a "canary" tag, especially since I'm testing a PR not a branch)

> I'll add some more custom error messaging for situations like this.

Did you do this yet? I'm still getting the same error:

```ShellSession
$ env -i PROTO_LOG=trace /Users/dan/.proto/shims/node --version
[TRACE 2023-12-15 17:01:21] proto_shim  Running node shim  shim="node" args=["/Users/dan/.proto/shims/node", "--version"] file="/Users/dan/.proto/shims/node"
[TRACE 01:21.613] proto_shim  Spawning proto child process  shim="node"
Error: No such file or directory (os error 2)
```

In particular, I don't know the variable environment when the shim is executed; nor for its call to `proto`.

FWIW, I don't use `$PROTO_INSTALL_DIR`, and I suspect the core issue here is that it's not running my login scripts which set up `$PROTO_HOME` and `$PATH`.

milesj:
@rotu Where is your `proto` located? If it's not on `PATH`, I'm not sure where else to find it. Already checking a handful of places.

rotu:
> @rotu Where is your `proto` located? If it's not on `PATH`, I'm not sure where else to find it. Already checking a handful of places.

My `proto` is in `/Users/dan/.proto/bin/proto`. I've been playing with symlinking it from `/Users/dan/.proto/shims/proto` so I don't accidentally circumvent the shims.

But I'm not sure how VSCode is invoking `node`, so it's not safe to assume that my zsh-specific `PATH`/`PROTO_HOME`/`PROTO_INSTALL_DIR` customizations are run!

rotu:
Okay I think I figured it out. `node --version` is hanging when invoked by VSCode. I can see this in VSCode's "Process Explorer" and the system "Activity Monitor". I'm pretty sure `node --version` should exit quickly but your comment [here](https://github.com/moonrepo/proto/blob/18378e0f0ecce15d103afd76ff6ecfb838269867/crates/cli/src/main_shim.rs#L187-L191) gave me an idea:

- This never terminates: `yes | /Users/dan/.proto/shims/node --version`
- This terminates quickly: `yes | /Users/dan/.proto/tools/node/20.10.0/bin/node --version`

I think the issue is similar - VSCode is communicating with the subprocess by some FIFO and the shim is inappropriately waiting for it to end.

Nothing exotic seems to be going on with `$PATH` or the variable environment, and my experiments with `env -i` are a red herring.

milesj:
Alright, good debugging. Let me see if there's a better way to handle piped data.

rotu:
I think simply ripping out the custom stream handling makes it do the right thing:
[stdio.patch](https://github.com/moonrepo/proto/files/13691884/stdio.patch)

milesj:
I had that originally and it didn't work correctly. Let me try testing stuff again now that it's more complete.

rotu:
The thing I was worried about there is that calling `spawn` might close the parent process's `stdin`, But that doesn't seem to be the case.

Also probably better not to test with `yes` - it produces a LOT of content! `cat -u - | node -i` or `cat -u sourcefile | pv --quiet --rate-limit 100 | node -i` is probably a better test setup.

I expect the shim should *not* change whether stdin is detected as a terminal. And `ctrl-c` should act as the original binary does - the shim should not transform the signal like it does now.

rotu:
> (fwiw, I think it's easier to have this be available on the PR action rather than as a release with a "canary" tag, especially since I'm testing a PR not a branch)

I went to test this again and I was seeing the same pipe issues. I then realized that I caught it in a state where only *some* of the binaries have been updated (I they all have been updated since).

I think I'm gonna have to PR the PR workflow :-p.

image

crates/cli/src/main_shim.rs
```diff
@@ -0,0 +1,211 @@
...
+    );
+
+    let shared_child = SharedChild::spawn(&mut command)?;
+    let child = Arc::new(shared_child);
+    let child_clone = Arc::clone(&child);
+
+    // Handle CTRL+C and kill the child
+    ctrlc::set_handler(move || {
+        trace!("Received ctrl + c, killing child process");
+        let _ = child_clone.kill();
```
rotu, line 201: I don't understand why this is necessary (at least on macOS and Linux). It has the adverse effect of altering the exit code of the subprocess from 130 to 1.

```
$ timeout --preserve-status --signal=INT 0.2 node; echo "exit code=$?"
...
exit code=1
$ timeout --preserve-status --signal=INT 0.2 $(proto bin node); echo "exit code=$?"
...
exit code=130
```

rotu: On the other hand, if the shim process receives a signal, but not the child, we aren't forwarding it to the child process.

I'm doing this with `timeout --foreground`. Can you tell that I'm having fun with my new sledgehammer?

e.g. Launching node directly and signaling it causes a graceful exit
```ShellSession
$ timeout --signal=QUIT --foreground 1 $(proto bin node)
Welcome to Node.js v20.10.0.
Type ".help" for more information.
>
$
```

e.g. This kills the shim but not `node`, causing `node` to subsequently crash:
```ShellSession
$ timeout --signal=QUIT --foreground 1 node
Welcome to Node.js v20.10.0.
Type ".help" for more information.
>
$
node:events:492                                                                 
      throw er; // Unhandled 'error' event
      ^

Error: read EIO
    at TTY.onStreamRead (node:internal/stream_base_commons:217:20)
    at TTY.callbackTrampoline (node:internal/async_hooks:130:17)
Emitted 'error' event on REPLServer instance at:
    at ReadStream.onerror (node:internal/readline/interface:239:10)
    at ReadStream.emit (node:events:514:28)
    at ReadStream.emit (node:domain:488:12)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -5,
  code: 'EIO',
  syscall: 'read'
}

Node.js v20.10.0
```

milesj: I've always been under the impression that ctrl+c'ing the parent process does not kill the child process, and it becomes a detached/zombie process. This has been a big problem in moon. https://doc.rust-lang.org/std/process/struct.Child.html

Let me test some more.

milesj: I think to solve this correctly, we somehow need to wrap these in a process group, at least on unix.

milesj: This is a good thread: https://github.com/rust-lang/cargo/issues/2343

rotu: I am well out of my depth here, but I don’t know that a process group is the thing to do here.

I know that if we *do* spawn a subprocess, that we can’t just let the default signal handlers bring down the parent process (hence why the ctrl-c handler helps - it suppresses the default behavior, but we need to deal with other signals like `SIGQUIT`).

If we don’t actually need to do anything after `node` ends, we can dodge the whole question by using `execve` on macOS and Linux. Maybe even [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/execve-wexecve) - I’m not sure what it actually does there, but I bet it’s close enough to work!

milesj: I tried 3 implementations, and only the `ctrlc` crate actually killed the child processes.

- Cargos impl: https://github.com/rust-lang/cargo/blob/master/crates/cargo-util/src/process_builder.rs
- https://crates.io/crates/ctrlc
- https://crates.io/crates/command-group

rotu: Looks like that cargo link was coming to the same conclusions I was moseying towards :-).

I particularly like `ProcessBuilder::exec_replace`, and I think it gives a simple effective implementation:

[exec_replace.patch](https://github.com/moonrepo/proto/files/13699301/exec_replace.patch)

> I tried 3 implementations, and only the ctrlc crate actually killed the child processes.

By design, ctrl-c sends an interrupt signal, which the process can choose to block, handle, or ignore. Depending on the test setup, you might have been checking the wrong behavior - the check should be that the child process was delivered the intended signal (and only if the signal caused the child process to end should the parent process stop supervising it).

milesj: I've been testing these outside of actually tests and running them manually.

I think another part of the problem is that there area actually 2 levels deep of child processes. That may be what's causing some confusion. It basically goes `node shim (1) -> proto run node (2) -> bin/node (3)`, and only 1 and 2 actually are handled correctly. Let me dig a further and maybe abstract all of this onto the same implementation.

milesj: Got a lot of this working here: https://github.com/moonrepo/proto/pull/346

rotu:
In testing this on Windows, I'm finding a broken `npm.cmd` and `npx.cmd` in `.proto/bin`. Namely, it has the line `SET "NPM_CLI_JS=%~dp0\node_modules\npm\bin\npm-cli.js"` (`%~dp0` expands to "the directory of the current script") and so the script fails at runtime with:

> Error: Cannot find module 'C:\Users\dan\.proto\bin\node_modules\npm\bin\npm-cli.js'

I think this is probably not a new issue but uncovered by the fact that the `npm.cmd` script is no longer shadowed by a `npm.cmd` shim.

I'm not directly calling `npm.cmd` but I think something in electron-builder is.

[npm.cmd.txt](https://github.com/moonrepo/proto/files/13709036/npm.cmd.txt)

milesj:
@rotu Nice find. I've run into that error before but wasn't sure what was causing it. I'll try and get a fix in for the 0.26 release.

milesj:
Posting a blog post about an RC to see if others want to help test: https://github.com/moonrepo/moon/pull/1236

rotu:
Just gave it another try on my Mac and played around with `kill`. There's still a bit of weirdness with signals but nothing major:

1. Running the node shim and sending either `node` or the shim a `kill -SIGINT` exits with return code 1 instead of 130 (which is what I get when running `node` directly).
2. Sending a `SIGSTOP` or `SIGTSTP` to the node process hangs the terminal.
3. Relatedly, sending a `SIGTSTP` by pressing ctrl-z apparently makes node nonresponsive. So you have to press ctrl-z a second time to actually get back to the terminal prompt (and then you can do `fg` to get back into the `node` repl). I'd expect to only have to ctrl-z once.
4. Sending the shim a `SIGALRM`, `SIGVTALRM`, `SIGPROF`, or `SIGPIPE` causes the shim to end but not bring down the `node` process (which may then crash with `Error: read EIO` if running interactively). Maybe something could be done with `libc::atexit` so `node` does not outlive the shim in these circumstances?

milesj:
@rotu Some of those signals we can't listen to: https://docs.rs/signal-hook/latest/signal_hook/low_level/fn.register.html#panics I'll try and listen on others. I don't think we'll reach 100% parity here.

As for exit codes, the ones explicitly set in a JS file is bubbled up correctly, like here: https://github.com/moonrepo/proto/blob/develop-0.26/docs/shim-test.mjs#L5 But whatever node is passing, I'm not sure why it doesn't grab it. The exit code is `None` on the rust side.

rotu:
> @rotu Some of those signals we can't listen to: https://docs.rs/signal-hook/latest/signal_hook/low_level/fn.register.html#panics I'll try and listen on others. I don't think we'll reach 100% parity here.

That's why I omitted `SIGKILL` :-). But that doesn't mean all is lost. For example, `SIGSTOP` hangs the terminal when sent to the `node` child process. In that case, the shim will get a `SIGCHLD` event and can act on it.

> As for exit codes, the ones explicitly set in a JS file is bubbled up correctly, like here: https://github.com/moonrepo/proto/blob/develop-0.26/docs/shim-test.mjs#L5 But whatever node is passing, I'm not sure why it doesn't grab it. The exit code is `None` on the rust side.

That's documented on [code()](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code):

> In Unix terms the return value is the *exit status*: the value passed to `exit`, if the process finished by calling `exit`. Note that on Unix the exit status is truncated to 8 bits, and that values that didn’t come from a program’s call to `exit` may be invented by the runtime system (often, for example, 255, 254, 127 or 126).
> 
> On Unix, this will return `None` if the process was terminated by a signal. [`ExitStatusExt`](https://doc.rust-lang.org/std/os/unix/process/trait.ExitStatusExt.html) is an extension trait for extracting any such signal, and other details, from the `ExitStatus`.

So I think you can check if:
- the child exit status has `stopped_signal` - I'm not sure what we need to do there to yield control back to the terminal; I guess wait for the `SIGCONT` somehow, maybe involving [`continued()`](https://doc.rust-lang.org/std/os/unix/process/trait.ExitStatusExt.html#tymethod.continued) which I don't really grok?
- the child exit status has `signal` (e.g. `SIGTERM`, `SIGKILL`, ...) - if so, end the shim with signal + 128.
- the child exit code has `code` and, if so, return it.

Of course, this is platform-specific and the juice might not be worth the squeeze, especially if it complicates the code too much!

milesj:
rc.3 will probably be the last release. I think this is good to go.

rotu:
LFGTM!!! :shipit:

PR #343, opened 2023-12-16
new: Add tests and handle piping.

milesj:


milesj:
@rotu Will land this and we can test some more.

Added unit tests and they seem to pass correctly.

rotu:
You may also want to do a test for handling of `SIGPIPE`. It seems like a rusty broken pipe could give you tetanus.

milesj:
@rotu I use this _which should_ avoid it: https://github.com/moonrepo/proto/blob/develop-0.26/crates/cli/src/main_shim.rs#L152

rotu:
> @rotu I use this _which should_ avoid it: https://github.com/moonrepo/proto/blob/develop-0.26/crates/cli/src/main_shim.rs#L152

I saw that. It would be good to have a test that *would fail* if that line were removed :-)

PR #344, opened 2023-12-17
Build and upload in CI

rotu:
This is a simpler workflow to publish binaries for release testing.
Associating the binaries with the CI workflow rather than a separate GitHub release has pros and cons.

Pro Artifacts:
- It's obvious which PR/commit the artifacts are associated with. When there are multiple commits, each one gets its own CI run and set of artifacts.
- Cluttered release page makes it hard to find a release you want.
- The CI results are available on the artifacts page, not so with releases.

Pro Release:
- Predictable URL better for linking
- Control over archives: can choose non-zip format, can specify file permissions (so no need to `chmod +x`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,11 +109,11 @@ jobs:
       - uses: moonrepo/setup-rust@v1
         with:
           targets: ${{ matrix.target }}
+          cache-target: release
       - run:
           cargo install --root artifacts --path ./crates/cli
       - uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.target }}
           path: artifacts
-          retention-days: 14
-          if-no-files-found: error
+          retention-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - if: ${{matrix.target == 'aarch64-unknown-linux-gnu'}}
-        run: apt install gcc-aarch64-linux-gnu
+        run: sudo apt-get install gcc-aarch64-linux-gnu
       - uses: moonrepo/setup-rust@v1
         with:
           targets: ${{ matrix.target }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,7 @@ jobs:
           targets: ${{ matrix.target }}
           cache-target: release
       - run:
-          cargo install --root artifacts --path ./crates/cli
+          cargo install --root artifacts --path ./crates/cli --target ${{ matrix.target }}
       - uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


milesj:
2 things come to mind:

- We need to use `--release`, otherwise there's a huge perf hit around WASM.
- Running this on every commit is overkill. Should probably just be commits to master/develop.

rotu:
> 2 things come to mind:
> 
> * We need to use `--release`, otherwise there's a huge perf hit around WASM.

I'm using `cargo install` which builds in release mode.

> * Running this on every commit is overkill. Should probably just be commits to master/develop.

This shouldn't be much bigger a hit than building for tests. Should I try to optimize the builds with some caching? Are you concerned about cpu time, storage, and/or something else?

---

Also can you comment on why the release workflows exist? It seems like a very different workflow for not a very different result.

.github/workflows/ci.yml
```diff
@@ -87,3 +87,33 @@ jobs:
...
+    strategy:
+      matrix:
+        include:
+        - target: aarch64-unknown-linux-gnu
+          host: ubuntu-latest
+        - target: x86_64-unknown-linux-gnu
+          host: ubuntu-latest
+        - target: x86_64-apple-darwin
+          host: macos-latest
+        - target: aarch64-apple-darwin
```
milesj, line 102: does this actually work? i dont think it does by default without extra configuration.

.github/workflows/ci.yml
```diff
@@ -87,3 +87,33 @@ jobs:
       - run: cargo run -- --help
       - run: cargo run -- list-remote node
       - run: cargo run -- list-remote wasm-test
+  build:
+    name: Upload build
+    runs-on: ${{ matrix.host }}
+    strategy:
+      matrix:
+        include:
+        - target: aarch64-unknown-linux-gnu
```
milesj, line 96: this is wrong? `cargo install` is probably installing the host target, which is x86_64

milesj:
I just don't see the need in building every commit. 99% of the time we don't care about it, and for those 1%, we can just enable the workflow to build and test what we need.

cargo-dist (https://opensource.axo.dev/cargo-dist/) does a lot of stuff that simple cargo build/install does not.

.github/workflows/ci.yml
rotu: D'oh! I left off the `--target` from `cargo install`

rotu (commented):
Build for the intended target (oops!)

.github/workflows/ci.yml
```diff
@@ -87,3 +87,36 @@ jobs:
...
+          host: ubuntu-latest
```
milesj, line 97: youll need the build jet runner for this to work

rotu: We should be able to cross-compile for aarch64 from x64.

I *did* need to install a suitable compiler with `sudo apt-get install gcc-aarch64-linux-gnu` and then it worked: https://github.com/rotu/proto/actions/runs/7279006762/job/19834420423?pr=1

PR #346, opened 2023-12-18
new: Improve signal handling.

milesj:
Uses `signal-hook` (which also has windows support) to handle signals correctly. I also wrapped the `proto run` in the same logic, since the flow is `shim -> proto run -> tool`.


rotu:
FYI, I'm currently checking out and testing PRs locally by using:

```
cargo install --git=https://github.com/moonrepo/proto --root=$env:PROTO_HOME --target-dir=$ENV:PROTO_HOME/build --rev=refs/pull/346/merge --force
```

This might be good enough to render redundant #344.

milesj:
Crazy, didnt even know that was possible, hah.

PR #350, opened 2023-12-19
new: Use `execvp` on Unix for better handling. Create new shim crate.

milesj:
I tried this before and it didn't work, but I think the problem was that I only applied it to the parent process, and not the intermediate process. So this applies it to both locations.


milesj:
@rotu

crates/cli/src/commands/run.rs
```diff
@@ -191,22 +201,13 @@ pub async fn run(args: ArgsRef<RunArgs>, proto: ResourceRef<ProtoResource>) -> S
...
-
     // Run after hook
-    if status.success() {
-        tool.run_hook("post_run", || RunHook {
-            context: tool.create_context(),
-            passthrough_args: args.passthrough.clone(),
-        })?;
-    }
+    // if status.success() {
+    //     tool.run_hook("post_run", || RunHook {
```
milesj, line 206: we lose this hook though, but i dont think its being used yet anyways

PR #351, opened 2023-12-19
Remove `version` from local crates

rotu:
Having both a `path` and a `version` causes Cargo to use a different package source when built locally versus from cargo.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -27,9 +27,9 @@ name = "proto"
 path = "src/main.rs"
 
 [dependencies]
-proto_core = { version = "0.25.2", path = "../core" }
-proto_pdk_api = { version = "0.11.2", path = "../pdk-api" }
-system_env = { version = "0.1.8", path = "../system-env" }
+proto_core = { path = "../core" }
+proto_pdk_api = { path = "../pdk-api" }
+system_env = { path = "../system-env" }
 chrono = "0.4.31"
 clap = { workspace = true, features = ["derive", "env"] }
 clap_complete = { workspace = true }
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -8,14 +8,14 @@ homepage = "https://moonrepo.dev/proto"
 repository = "https://github.com/moonrepo/proto"
 
 [dependencies]
-proto_pdk_api = { version = "0.11.2", path = "../pdk-api" }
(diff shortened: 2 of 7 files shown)


rotu:
I'm not sure if this is the desired solution. My intent is to fix this build error when I try installing with Cargo:
```ShellSession
$ cargo install --git=https://github.com/moonrepo/proto --root=~/.proto --target-dir=~/.proto/build --force --locked
...
   Compiling system_env v0.1.8 (/Users/dan/.cargo/git/checkouts/proto-2c037be0d9f54cdf/2e8260e/crates/system-env)
error[E0599]: no method named `as_encoded_bytes` found for reference `&OsStr` in the current scope
  --> crates/system-env/src/helpers.rs:85:10
   |
84 |       let bin_path = if bin
   |  _______________________-
85 | |         .as_encoded_bytes()
   | |         -^^^^^^^^^^^^^^^^ method not found in `&OsStr`
   | |_________|
   | 

error[E0599]: no method named `as_encoded_bytes` found for reference `&OsStr` in the current scope
   --> crates/system-env/src/helpers.rs:104:61
    |
104 | ...a.as_ref().as_encoded_bytes()).to_string())
    |               ^^^^^^^^^^^^^^^^ method not found in `&OsStr`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `system_env` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: failed to compile `proto_cli v0.25.3 (https://github.com/moonrepo/proto#2e8260ef)`, intermediate artifacts can be found at `/Users/dan/~/.proto/build`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

milesj:
Publishing to crates.io fails if there's no `version` field -_-

rotu:
> My intent is to fix this build error when I try installing with Cargo on macOS.

Totally misdiagnosed this issue. I think the cause is that [`as_encoded_bytes`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.as_encoded_bytes) is new in 1.74.0 but I was building with cargo 1.73.0.

rotu:
> Publishing to crates.io fails if there's no `version` field -_-

Ah that makes sense. Closing.

PR #352, opened 2023-12-20
eschew as_encoded_bytes

rotu:
We were abusing it anyway. We were looking at the raw bytes and interpreting as utf8 instead of using the OS encoding.
After this, we can build on rust as far back as 1.70.0 (dependency on `anstyle` declares that as a minimum rustc)


rotu:
This also fixes the build under cargo 1.73.0 mentioned [here](https://github.com/moonrepo/proto/pull/351#issuecomment-1863624489):
```ShellSession
$ cargo install --git=https://github.com/moonrepo/proto --root=~/.proto --target-dir=~/.proto/build --force --locked
...
   Compiling system_env v0.1.8 (/Users/dan/.cargo/git/checkouts/proto-2c037be0d9f54cdf/2e8260e/crates/system-env)
error[E0599]: no method named `as_encoded_bytes` found for reference `&OsStr` in the current scope
  --> crates/system-env/src/helpers.rs:85:10
   |
84 |       let bin_path = if bin
   |  _______________________-
85 | |         .as_encoded_bytes()
   | |         -^^^^^^^^^^^^^^^^ method not found in `&OsStr`
   | |_________|
   | 

error[E0599]: no method named `as_encoded_bytes` found for reference `&OsStr` in the current scope
   --> crates/system-env/src/helpers.rs:104:61
    |
104 | ...a.as_ref().as_encoded_bytes()).to_string())
    |               ^^^^^^^^^^^^^^^^ method not found in `&OsStr`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `system_env` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: failed to compile `proto_cli v0.25.3 (https://github.com/moonrepo/proto#2e8260ef)`, intermediate artifacts can be found at `/Users/dan/~/.proto/build`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

crates/system-env/src/helpers.rs
```diff
@@ -14,10 +14,7 @@ pub fn find_command_on_path<T: AsRef<OsStr>>(name: T) -> Option<PathBuf> {
     // Only extensions we care about
     let exts = vec![".exe", ".ps1", ".cmd", ".bat"];
     let name = name.as_ref();
-    let has_ext = name
-        .as_encoded_bytes()
```
milesj, line 18: I originally went with this to avoid unnecessary string conversion.

rotu: I think it's wrong in the case of UTF-16, where you're going through bytes instead of characters.

But also, the logic here is kinda iffy - on Windows, `PATHEXT` is appended to executable names. So `foo.bar` will look for `foo.bar.exe`, even though there is an extension there. Maybe we should just change it to `let exts = vec![".exe", ".ps1", ".cmd", ".bat", ""];` and skip the conditional?

PR #365, opened 2023-12-27
new: Support PowerShell profiles.

milesj:


milesj:
@rotu Is this how you use pwsh?

rotu:
> @rotu Is this how you use pwsh?

Roughly, yes.
1. I only use `Documents\PowerShell\Profile.ps1`, not `Microsoft.PowerShell_profile.ps1`. For example, the latter is not respected by PowerShell running in VSCode.
2. You may want to use `[IO.Path]::PathSeparator` instead of a literal `;` or `:`.
3. I think `env::var("PSHOME")` is not right. Powershell sets `$PSHOME`, and I think you’re looking at `$env:PSHOME`.

milesj:
1/3 - Basing on this doc: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.4#profile-types-and-locations

2 - I may be blind, but I couldn't find a constant in Rust std for the path delimiter.

rotu:
> 1/3 - Basing on this doc: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.4#profile-types-and-locations

Careful. “Host” there means “Application hosting the shell”, not “host machine”

Also, in PowerShell, `$PSHOME` and `$HOME` are variables but *not* environment variables. `$env:PSHOME` and `$env:HOME` *are* environment variables. I would not expect them to be set (maybe they are but that page doesn’t tell me)

> 2 - I may be blind, but I couldn't find a constant in Rust std for the path delimiter.

I’m saying don’t split the Windows and non-Windows cases. Leave is as `[IO.Path]::PathSeparator` in the shell script and let PowerShell resolve this to the particular character. https://powershell.org/2019/02/tips-for-writing-cross-platform-powershell-code/

milesj:
> Careful. “Host” there means “Application hosting the shell”, not “host machine”
> Also, in PowerShell, $PSHOME and $HOME are variables but not environment variables. $env:PSHOME and $env:HOME are environment variables. I would not expect them to be set (maybe they are but that page doesn’t tell me)

Well that's super confusing. I don't see any clarification in the docs either, but not supporting `PSHOME` is probably fine.

> I’m saying don’t split the Windows and non-Windows cases. Leave is as [IO.Path]::PathSeparator in the shell script and let PowerShell resolve this to the particular character. https://powershell.org/2019/02/tips-for-writing-cross-platform-powershell-code/

Ah in the PS code, thought you meant Rust side.

rotu:
This is what the relevant part of my PowerShell profile looks like:

```ps
$env:PROTO_HOME = Join-Path $HOME '.proto'
$env:PATH = @(
    (Join-Path $env:PROTO_HOME 'shims'), 
    (Join-Path $env:PROTO_HOME 'bin'),
    $env:PATH
) -join [IO.PATH]::PathSeparator

$env:PATH = ($env:PATH.split([IO.Path]::PathSeparator) | Select-Object -Unique) -join [IO.PATH]::PathSeparator
```

milesj:
@rotu Thanks thats super useful. Updated.

Issue #361, opened 2023-12-26
Setup for Elvish shell is not working as intended


frantic1048:
### What version?

0.26.0

### Which command?

curl -fsSL https://moonrepo.dev/install/proto.sh | bash

## Problem

**Elvish version**: 0.19.2

After running installation script:

```sh
curl -fsSL https://moonrepo.dev/install/proto.sh | bash
```

It appends following code to `~/.config/elvish/rc.elv`

```elvish
# proto
set-env PROTO_HOME $HOME/.proto
set-env PATH (str:join ':' [$PROTO_HOME/shims:$PROTO_HOME/bin $E:PATH])
```

This will cause the following error:

```
Multiple compilation errors in /Users/kafuuchino/.config/elvish/rc.elv:
  variable $HOME not found
    27:20: set-env PROTO_HOME $HOME/.proto
  variable $PROTO_HOME not found
    28:29: set-env PATH (str:join ':' [$PROTO_HOME/shims:$PROTO_HOME/bin $E:PATH])
  variable $PROTO_HOME not found
    28:47: set-env PATH (str:join ':' [$PROTO_HOME/shims:$PROTO_HOME/bin $E:PATH])
```

The correct code could be:

```elvish
# proto

# PROTO_HOME
set-env PROTO_HOME {~}/.proto

# PATH

# via string concat, add the missing `E:` prefix
set-env $PATH (str:join ':' [
  $E:PROTO_HOME/shims
  $E:PROTO_HOME/bin
  $E:PATH
])

# or via special paths variable for easier edit
set paths = [
  $E:PROTO_HOME/shims
  $E:PROTO_HOME/bin
  $@paths
]
```

## Explanation

To reference an environment variable named `PROTO_HOME`, either prefix with `E:` or use `get-env` builtin:
- not work: `echo $PROTO_HOME`: this is a reference to local variable named `PROTO_HOME`, not environment variable.
- works: `echo $E:PROTO_HOME` or `echo (get-env PROTO_HOME)`

## References

- dedicated `E:` namespace for environment variables: https://elv.sh/ref/language.html#special-namespaces
- `get-env`: https://elv.sh/ref/builtin.html#get-env
- Special `$paths` builtin variable for easier operating of `$E:PATH`, `$paths` is a list form of `$E:PATH`, allowing list operations to be directly applied to `$E:PATH`: https://elv.sh/ref/builtin.html#$paths

### Operating system?

MacOS

### Architecture?

x64

[...]

PR #366, opened 2023-12-27
refactor: refine Elvish profile setup

frantic1048:
1. Remove `str:join` usage for better portability and simplicity

   using `str:join` requires `use str` first, and we can directly operate on `$paths` to avoid dealing with `:` in `PATH` manually

2. Add general logic to deal with normal env vars

relates to https://github.com/moonrepo/proto/issues/361
follows https://github.com/moonrepo/proto/pull/362

---

I'm very new to Rust and I'm afraid if I have done anything wrong here. Please point out if any exists :)

### DIFF
--- a/crates/cli/src/shell.rs
+++ b/crates/cli/src/shell.rs
@@ -77,20 +77,20 @@ pub fn format_env_var(shell: &Shell, key: &str, value: &str) -> Option<String> {
         } else {
             format!(r#"export {key}="{value}""#)
         }),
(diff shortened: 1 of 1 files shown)


frantic1048:
Replacing `$E:HOME` with `{~}` does not change the meaning of code.

For completeness and readability of output, I added the `replace()` to both `PATH` and other env vars.

milesj:
Awesome thanks!

PR #456, opened 2024-04-06
new: Automatically configure Windows system PATH

Phault:
Very much inspired by rustup's implementation and in fact the low-level byte handling was copied verbatim, hence I added their MIT license. I hope this is fine with you, as I don't really know another way to achieve that part.

I added a --no-modify-path parameter to the setup command, but it's only really relevant for Windows. Should I conditionally compile it in, or is it better to keep the API identical?

I also had to restructure the setup flow quite a bit. I think I kept the same spirit of it, but let me know if you're unhappy with any of it. 

Note in the finished_message I opted for not suggesting Windows users to configure their PowerShell profile manually, and instead just tell them what to add to the system PATH if they didn't let us do it for them.

---

A final thing that just hit me, is something I encountered personally when I started using proto. I prepended my path (the user one) manually, just like this would do and for the most part it worked fine. Down the line I had an issue with a binary (dotnet maybe?) that was installed and added by its own installer to the system path (if I didn't just forget about doing it myself that is). This will always take precedence over the user one, as the user path is append to the system path. I ended up having to remove that binary from the system path. 

I don't see a way around this though... Maybe just something for the upcoming `proto doctor` command to check for and warn about?

### DIFF
--- a/crates/cli/src/commands/setup.rs
+++ b/crates/cli/src/commands/setup.rs
@@ -135,6 +135,27 @@ fn already_setup_message(installed_bin_path: PathBuf) {
     help_message();
 }
 
+fn manual_system_path_message(proto: &ProtoResource) {
+    println!(
+        "Add the following to your {} to get started:",
+        color::property("PATH")
+    );
+    println!();
+    println!(
+        "{}",
+        if cfg!(windows) {
+            // we avoid %USERPROFILE% as it only works in the user path and not system path
+            color::muted_light(format!(
+                "{};{}",
+                proto.env.store.shims_dir.as_os_str().to_string_lossy(),
+                proto.env.store.bin_dir.as_os_str().to_string_lossy()
+            ))
+        } else {
+            color::muted_light("$HOME/.proto/shims;$HOME/.proto/bin")
+        },
+    );
+}
+
 fn finished_message(
     proto: &ProtoResource,
     exported_content: Option<String>,
@@ -149,29 +170,18 @@ fn finished_message(
 
     if path_was_updated {
         println!("Launch a new terminal window to start using proto!");
-    } else if exported_content.is_some() && cfg!(not(windows)) {
-        println!("Add the following to your shell profile to get started:");
+    } else if exported_content.is_some() {
+        if cfg!(windows) {
+            manual_system_path_message(proto);
+            println!();
+            println!("Or alternatively add the following to your shell profile:");
+        } else {
+            println!("Add the following to your shell profile to get started:");
+        }
         println!();
         println!("{}", color::muted_light(exported_content.unwrap().trim()));
     } else {
-        println!(
-            "Add the following to your {} to get started:",
-            color::property("PATH")
-        );
-        println!();
-        println!(
-            "{}",
-            if cfg!(windows) {
-                // we avoid %USERPROFILE% as it only works in the user path and not system path
-                color::muted_light(format!(
-                    "{};{}",
-                    proto.env.store.shims_dir.as_os_str().to_string_lossy(),
-                    proto.env.store.bin_dir.as_os_str().to_string_lossy()
-                ))
-            } else {
-                color::muted_light("$HOME/.proto/shims;$HOME/.proto/bin")
-            },
-        );
+        manual_system_path_message(&proto);
     }
 
     println!();
--- a/crates/cli/src/commands/setup.rs
+++ b/crates/cli/src/commands/setup.rs
@@ -52,16 +52,6 @@ pub async fn setup(args: ArgsRef<SetupArgs>, proto: ResourceRef<ProtoResource>) 
 
     println!("Finishing proto installation...");
 
-    let mut path_was_updated = false;
-
-    #[cfg(windows)]
-    if !args.no_modify_path {
-        path_was_updated |= windows::do_add_to_path(vec![
-            proto.env.store.shims_dir.clone(),
-            proto.env.store.bin_dir.clone(),
-        ])?;
-    }
-
     let content = format_exports(
         &shell,
         "proto",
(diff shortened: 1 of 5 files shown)


crates/cli/src/commands/setup/windows.rs
```diff
@@ -0,0 +1,154 @@
...
+        })
+        .into_diagnostic()
+}
+
+fn set_path_var(new_path: &[u16]) -> Result<()> {
+    RegKey::predef(HKEY_CURRENT_USER)
+        .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
+        .and_then(|environment| {
+            if new_path.is_empty() {
+                environment.delete_value("PATH")
```
milesj, line 62: Does this delete the whole value? Seems risky no?

Phault: Yeah it does, but if removed it would just be emptying the variable instead. Same effective result.

If you ever plan to have an uninstaller, then this would make sense to keep. But for now it's probably better I just make it error if the new_path is empty.

crates/cli/src/commands/setup.rs
```diff
@@ -154,38 +136,42 @@ fn already_setup_message(installed_bin_path: PathBuf) {
...
 fn finished_message(
-    installed_bin_path: PathBuf,
-    updated_profile_path: Option<PathBuf>,
+    proto: &ProtoResource,
     exported_content: Option<String>,
+    path_was_updated: bool,
 ) {
-    if let Some(profile_path) = updated_profile_path {
-        println!(
-            "Successfully installed proto to {} and updated profile {}!",
```
milesj, line 163: Letting users know which profile was updated is very important for users, since they don't always choose it.

crates/cli/src/commands/setup.rs
```diff
@@ -154,38 +136,42 @@ fn already_setup_message(installed_bin_path: PathBuf) {
...
+    let installed_bin_path = proto.env.store.bin_dir.join(get_exe_file_name("proto"));
+
+    println!(
+        "Successfully installed proto to {}!",
+        color::path(installed_bin_path),
+    );
+
+    if path_was_updated {
         println!("Launch a new terminal window to start using proto!");
+    } else if exported_content.is_some() && cfg!(not(windows)) {
```
milesj, line 152: What about the content & windows case? We can't assume 100% of windows users would want to edit system vars.

crates/cli/src/commands/setup.rs
Phault: Yeah, it was just moved to https://github.com/moonrepo/proto/pull/456/files#diff-a31902d06ac6afdddf64feb76dc589f59e39a33434c352de60ba84f8beca4427R116 instead

crates/cli/src/commands/setup.rs
```diff
@@ -132,10 +113,11 @@ pub async fn setup(args: ArgsRef<SetupArgs>, proto: ResourceRef<ProtoResource>)
 
     // If we found a profile, update the global config so we can reference it
     if let Some(profile) = &profile_path {
+        println!("Updated profile {}", color::path(profile));
```
milesj, line 116: This would appear before the success messages right? Feels a bit weird.

crates/cli/src/commands/setup.rs
Phault: This is what I was referring to here:

> Note in the finished_message I opted for not suggesting Windows users to configure their PowerShell profile manually, and instead just tell them what to add to the system PATH if they didn't let us do it for them.

Essentially I just think it's far more likely a Windows user will want to update their system PATH as that is effectively what all other programs do. Especially since that's often the only way it'll work with their IDEs.

I can change it back if you'd like.

crates/cli/src/commands/setup.rs
Phault: Yup. 

I dunno, not weird to me at least. Isn't that usually what CLIs do, log progress as they go?

crates/cli/src/commands/setup/windows.rs
milesj: Yeah it just seems odd. When would `new_path` be empty? Since we own that value and pass it in.

crates/cli/src/commands/setup.rs
milesj: Yeah I'm not sure which is best. I wish there was stats on what users typically do, as I've heard both arguments.

Maybe we show both messages so they can choose?

Phault: Yeah that sounds preferable. I'll update it to look like this:

```shell
PS D:\Projects\proto> .\target\debug\proto.exe setup --no-modify-path --no-profile
Finishing proto installation...
Successfully installed proto to ~\.proto\bin\proto.exe!
Add the following to your PATH to get started:

C:\Users\caspe\.proto\shims;C:\Users\caspe\.proto\bin

Or alternatively add the following to your shell profile:

# proto
$env:PROTO_HOME = Join-Path $HOME ".proto"
$env:PATH = @(
  (Join-Path $env:PROTO_HOME "shims"),
  (Join-Path $env:PROTO_HOME "bin"),
  $env:PATH
) -join [IO.PATH]::PathSeparator

Need help? Join our Discord https://discord.gg/qCh9MEynv2
```

Unsure about the phrasing, so just shout if you want it changed :)

crates/cli/src/commands/setup/windows.rs
Phault: True. Replaced that check with an assertion that it isn't empty.

crates/cli/src/commands/setup.rs
Phault: Moved it back to finished_message

crates/cli/src/commands/setup.rs
Phault: Just gonna trust your gut on this. I've moved it back, although its still on its own line after the success message, due to all the message variations now. 

The finished_message has gotten a bit messy, if you have an idea for restructuring the function I'm all ears :)

milesj:
Thanks, great start! Will merge and test on my machine.

PR #463, opened 2024-04-15
new: use .zshenv by default

latin-1:
**References:**
https://zsh.sourceforge.io/Intro/intro_3.html
> `.zshenv' is sourced on all invocations of the shell, unless the -f option is set. It should contain commands to set the command search path, plus other important environment variables.

**Implementation in rustup:**
https://github.com/rust-lang/rustup/blob/f26d16b3b940bfb7b5e5aad7ee8f7121569f1d60/src/cli/self_update/shell.rs#L186-L191


milesj (approved):

milesj:
Thanks!

Issue #483, opened 2024-05-28
Inhancement of node.js version detections


tolluset:
### What version?

0.35.3

### What happened?

When I install node modules packages, I met this message. 

```
Error: plugin::call_func::failed

  × unexpected character 'v' while parsing major version number 
```

The project using package.json for pin node version like this.

```
...

  "engines": {
    "node": ">=v18.0.0"
  }
```

The `v` makes incorrect. After I remove `v` works fine.

I think may support start with `v` node version would be more nice for compatibility to other tools?

### Operating system?

MacOS

### Architecture?

arm64

milesj:
This is reasonable, will look into.

PR #485, opened 2024-05-28
fix: recognize version with start with v when leading inequality sign

tolluset:
Refer to https://github.com/moonrepo/proto/issues/483 issue, some version using `v` version with inequality sign, So recognize that versions.

---

```
just test-ci

     Summary [ 259.601s] 233 tests run: 233 passed, 0 skipped
```


milesj:
Thanks for this, the failing tests are unrelated to your change.

PR #534, opened 2024-07-02
feat: add --include-global support to proto use

### LINKED ISSUE #533: Add `--include-global` support to `proto use`
Right now `proto use` only installs tools from the current working directory and upwards, but does not include the global `~/.proto/prototools`. We should add an argument to also include global versions.

Work:
- Update https://github.com/moonrepo/proto/blob/master/crates/cli/src/commands/install_all.rs
- Add arguments
- Rework line 24 (refer to outdated for an example)

varshith257:
Fixes #533 
/claim #533

### DIFF
--- a/crates/cli/src/commands/install_all.rs
+++ b/crates/cli/src/commands/install_all.rs
@@ -107,3 +107,39 @@ pub async fn install_all(session: ProtoSession, args: InstallAllArgs) -> AppResu
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::app::App;
+    use crate::session::ProtoSession;
+    use starbase::AppResult;
+
+    #[tokio::test]
+    async fn test_install_all_with_global() -> AppResult {
+        let app = App::new();
+        let session = ProtoSession::new(app);
+        let args = InstallAllArgs {
+            include_global: true,
+        };
+
+        let result = install_all(session, args).await;
+        assert!(result.is_ok());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_install_all_without_global() -> AppResult {
+        let app = App::new();
+        let session = ProtoSession::new(app);
+        let args = InstallAllArgs {
+            include_global: false,
+        };
+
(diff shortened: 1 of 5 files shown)


varshith257:
@milesj I have added changes to it. Pleasee review it and any changes required will incorporate them

crates/cli/src/main.rs
```diff
@@ -101,7 +101,7 @@ async fn main() -> MainResult {
             Commands::Uninstall(args) => commands::uninstall(session, args).await,
             Commands::Unpin(args) => commands::unpin(session, args).await,
             Commands::Upgrade => commands::upgrade(session).await,
-            Commands::Use => commands::install_all(session).await,
+            Commands::Use => commands::install_all(session, args).await,
```
milesj, line 104: You'll need to update the `Use` variant also.

milesj:
Can you add a test also.

crates/cli/src/main.rs
varshith257: Ok

varshith257:
I will add it

crates/cli/src/commands/install_all.rs
```diff
@@ -96,3 +107,39 @@ pub async fn install_all(session: ProtoSession) -> AppResult {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::app::App;
+    use crate::session::ProtoSession;
+    use starbase::AppResult;
+
+    #[tokio::test]
+    async fn test_install_all_with_global() -> AppResult {
+        let app = App::new();
```
varshith257, line 120: @milesj I am not sure about how we initialize app and test proto.

Ahh, any suggestions you incude here? Are tests is in right direction? I think need to explore proto more( IMO will get experience more on it by contributing more to proto)

milesj: @varshith257 We use integration tests for these, here's some refs: https://github.com/moonrepo/proto/blob/master/crates/cli/tests/use_test.rs

varshith257:
@milesj Aded integration tests. Have a look at it

PR #540, opened 2024-07-09
feat: support asdf plugins through a proto WASM plugin

### LINKED ISSUE #539: Support asdf plugins through a proto WASM plugin
There has been multiple requests now to support (piggy-back) off of [asdf](https://asdf-vm.com/), since it has a ton of plugins and support for many tools. Unfortunately, building this into proto core simply isn't feasible without rewriting it from scratch, and at that point, it's really not worth it. The other issue is that asdf is Unix only, so that means no Windows support.

However! I had a thought the other day, why not support _every asdf plugin_ through a single proto WASM plugin, similar to how our TOML plugins work (they are WASM under the hood). For example, say you wanted to use the [scala](https://github.com/asdf-vm/asdf-plugins/blob/master/plugins/scala) asdf plugin, it may look something like the following in proto:

```toml
# .prototools
scala = "latest"

[plugins]
scala = "https://path/to/asdf/plugin.wasm"
```

Easy right? But what if the proto identifier doesn't match up with the asdf plugin name? Well, just override it with configuration. For example, using the [nodejs](https://github.com/asdf-vm/asdf-plugins/blob/master/plugins/nodejs) asdf plugin.

```toml
# .prototools
node = "latest"

[plugins]
node = "https://path/to/asdf/plugin.wasm"

[tools.node]
asdf-plugin = "nodejs"
```

We can also take this further by supporting the git repository directly.

```toml
# .prototools
[tools.node]
asdf-repository = "https://github.com/asdf-vm/asdf-nodejs.git"
```

## Caveats

With all this being said, I have _no idea_ if this will actually work, or if the asdf plugin APIs will match up to the proto plugin APIs. It's definitely something that needs to be prototyped to ensure it's feasible.

## Requirements

- Plugin should be written in Rust and use our APIs/PDKs:
  - https://moonrepo.dev/docs/guides/wasm-plugins
  - https://moonrepo.dev/docs/proto/wasm-plugin
- PRs submitted to https://github.com/moonrepo/asdf-plugin
- Does _not_ need to support Windows
- Must work _without_ asdf existing in the environment
- Should ideally support all asdf plugins
- Should detect versions from `.tool-versions` - https://asdf-vm.com/manage/configuration.html#tool-versions
- Should support `ASDF_*` env vars - https://asdf-vm.com/plugins/create.html#environment-variables-overview

## Implementation

How this is implemented may require a lot of discussion, which can be had on Discord. 

With that said, we can outline a few things:

- How we download/install asdf plugins should probably mirror what asdf itself does.
  - Stored at `~/.asdf` - https://asdf-vm.com/manage/configuration.html#asdf-data-dir
- API reference - https://asdf-vm.com/plugins/create.html#required-scripts

varshith257:
Fixes #539 
/claim #539


crates/asdf-plugin/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+// unpack_archive: Unpacks downloaded archives based on their file extension.
+#[plugin_fn]
+pub fn unpack_archive(Json(input): Json<UnpackArchiveInput>) -> FnResult<()> {
+    let input_file = input.input_file;
+    let output_dir = input.output_dir;
+
+    // Need to ensure file type and unpack accordingly
+    if input_file.ends_with(".tar.xz") {
+        //TODO: Implement the untar and unzip moments
+        // untar(input_file, output_dir)?;
```
varshith257, line 175: Is this untar and unzip could be custom implemented or any suggested crates for it?

crates/asdf-plugin/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    Ok(())
+}
+
+// detect_version_files: Specifies which files to check for version information.
+#[plugin_fn]
+pub fn detect_version_files(_: ()) -> FnResult<Json<DetectVersionOutput>> {
+    let asdf_plugin = AsdfPlugin {
+        tool: Tool::default(),
```
varshith257, line 191: There is no default method for initializing an instance of AsdfPlugin with a default instance of Tool. Can it be implemented in Tool impl?

crates/asdf-plugin/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+        std::process::Command::new("git")
+            .arg("clone")
+            .arg(repository)
+            .arg(&plugin_dir)
+            .output()
+            .expect("Failed to clone asdf plugin");
+    }
+
+    pub fn pre_install(&self, mut input: InstallHook) {
+        let config: AsdfConfig = self.tool.config();
```
varshith257, line 119: I had confusion here about referencing config. Any guidance on this?

varshith257 (commented):
@milesj I have referenced above with my doubts. PTAL
( I have kept them as default values to better know their functionality there)

milesj:
@varshith257 Plugins don't belong in this repo, so it should probably be its own repo? I'm not sure how bounties work in that case.

crates/asdf-plugin/src/lib.rs
milesj: Use https://docs.rs/proto_pdk/latest/proto_pdk/fn.get_tool_config.html

varshith257:
@milesj We can implement it and reference this issue there and if a PR merged there it will automatically considered for claim and another approach is to maybe we can update our readme or docs right in that PR you can tip it .

varshith257:
Moving this work to the new repo as we discussed

Issue #711, opened 2025-02-05
Feature: New version tokens for TOML, JSON, and YAML plugins


tom-fletcher:
## Issue

Follow on from #710.

This relates only to TOML, JSON, and YAML plugins.

Some tools have non-standard versioning. For example, the `backend-convex` package releases daily builds tagged in the format:

```
precompiled-2025-02-04-7d6e1db
```

This can be resolved using the `resolve.version-pattern`:

```toml
[resolve]
git-url = "https://github.com/get-convex/convex-backend"
version-pattern = "^precompiled-(?<major>\\d{4})-(?<minor>\\d{2})-(?<patch>\\d{2})-(?<pre>[a-f0-9]+)$"
```

Which then allows the user to pin the version in their `.prototools` file as if it were `semver`:

```
convex-backend = "2025.02.04-7d6e1db"
```

_Note that calver isn't used here, due to the limitations discussed in #710 i.e. if `pre` starts with a number then calver can't validate the version_

All of this works fine, until the `download-url` is composed. This must be in the format:

```
https://github.com/get-convex/convex-backend/releases/download/precompiled-2025-02-04-7d6e1db/{download_file}"
```

However, `proto` doesn't currently provide granularity in the tokens to format this. The closest we can currently get is using `{version}`, which only allows us to create an invalid link as follows:

```
https://github.com/get-convex/convex-backend/releases/download/precompiled-2025.02.04-7d6e1db/{download_file}
```

## Solution

Adding additional version tokens allows the `download-link` to be composed more granularly. With `{versionMinor}` and `{versionPatch}`, this could be composed as follows:

```toml
[install]
download-url = "https://github.com/get-convex/convex-backend/releases/download/precompiled-{versionMajor}-{versionMinor}-{versionPatch}-{versionPrerelease}/{download_file}"
```

The `{versionMajorMinor}` and `{versionYearMonth}` tokens can also be deprecated at this stage, as if the user needs these, they can build these manually i.e. `{versionMajor}.{versionMonth}`.

## Pull requests

This requires changes (in the following order) to the [moonrepo/plugins](https://github.com/moonrepo/plugins), [moonrepo/proto](https://github.com/moonrepo/proto), and [moonrepo/moon](https://github.com/moonrepo/moon) repositories to update the `schema_tool`, version of `schema_tool` used by `proto`, and the website.

PRs are incoming, and will be linked to this issue shortly for tracking. 😄

tom-fletcher:
[...]

PR #712, opened 2025-02-05
deps: Update schema tool version

tom-fletcher:
See moonrepo/proto#711

This is dependant on moonrepo/plugins#34 first being released, and assumes the released version of `schema_tool` is `0.16.4`. If this changes, this PR also needs updated.

*I have assumed this will be released as 0.45.2 when bumping the version - if not then this will affect moonrepo/moon#1820*


milesj:
Ty!

PR #737, opened 2025-03-03
fix: use GitHub Releases API instead of raw version file

liby:
This change makes the version check more reliable by ensuring we only get versions that have complete release artifacts available. This prevents users from attempting to upgrade to versions where the binaries are not yet published.

### DIFF
--- a/crates/cli/src/helpers.rs
+++ b/crates/cli/src/helpers.rs
@@ -2,6 +2,7 @@ use miette::IntoDiagnostic;
 use semver::Version;
 use starbase_console::ui::{ConsoleTheme, Style, style_to_color};
 use starbase_styles::color;
+use starbase_utils::json::JsonValue;
 use tracing::debug;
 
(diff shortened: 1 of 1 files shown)


milesj:
The failing tests are interesting. I'm guessing the json decode is failing? Is the endpoint returning non-json by chance?

liby:
> The failing tests are interesting. I'm guessing the json decode is failing? Is the endpoint returning non-json by chance?

Yes, the issue was with missing User-Agent header. [GitHub API requires it](https://docs.github.com/en/rest/using-the-rest-api/getting-started-with-the-rest-api?apiVersion=2022-11-28#headers:~:text=Requests%20with%20no%20User%2DAgent%20header%20will%20be%20rejected), otherwise returns a 403 error instead of JSON. Adding the header fixed it.

milesj:
Thanks!

Issue #785, opened 2025-05-22
Prevent shims from blocking native binaries


RobertVillalba:
I have a repo where multiple teams are working together. One of the teams natively works in Golang and I happen to need to install go in order to build a binary as part of my project. I want to use proto to install and run go on my end but this means that a shim is created and once folks are outside of my folder it messes with their environment (even for myself). 

```
Error: proto::detect::failed

  × Failed to detect an applicable version to run Go with. Try pinning a version with proto pin or explicitly
  │ passing the version as an argument or environment variable.
  help: Has the tool been installed?
```

The issue is that my folder has go pinned in the `.prototools` file. When they exit the folder, the call to go hits the shim, doesn't see a pinned go version and returns that error. It would be great if the shim could see that there is no pinned version and fallback to calling the binary that would have been called if proto was not in the path. 

An alternative would be to force every team to only install using proto, but that does not seem viable or correct. We could use `activate` or something similar, but this would be a less ideal workflow and maybe still runs into similar issues?

I've seem slightly related mentions in a few places like here https://github.com/moonrepo/proto/issues/739#issuecomment-2704316632

Really loving the tool and the whole concept behind it. I'm trying to find ways to integrate it into our own workflow while not disrupting others and possibly introducing it to other teams as we move forward.

RobertVillalba:
I think the issue is that the `locate_tool` function finds go as a built in plugin and so it doesn't flag it as an unknown tool. I'm trying to add a check to ignore the built in tools if we are running from within a shim since I think this might be a corner case where we wouldn't be trying to auto install or anything like that. I don't know what all the possible scenarios might be though.

RobertVillalba:
@milesj I've updated the PR. I don't think I understood the structure enough the first go round. If you could please rerun the CI tests and see if the proposed solution seems acceptable to you.

RobertVillalba:
Perfect thank you!

PR #788, opened 2025-05-23
fix: run_global_tool

RobertVillalba:
Addresses issue in: https://github.com/moonrepo/proto/issues/785

It seems that `run_global_tool` was not correctly catching the case a left over shim from some project was present, but there was no globally pinned version. When the user attempted to run the native binary proto would automatically detect that the plugin was one of the built in ones and run through the regular process likely to provide the auto install option. I updated `locate_tool` to ignore the built-in plugins check if it detects that it is running from a shim since a shim should imply that the user should have a pinned version in one of the config files.

### DIFF
--- a/crates/cli/src/commands/run.rs
+++ b/crates/cli/src/commands/run.rs
@@ -3,6 +3,7 @@ use crate::error::ProtoCliError;
 use crate::session::ProtoSession;
 use clap::Args;
 use miette::IntoDiagnostic;
+use proto_core::flow::detect::ProtoDetectError;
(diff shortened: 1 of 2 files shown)


RobertVillalba:
I've pushed the test fixes as well now.

milesj:
Yeah these changes make sense. Let's see what the tests say.

milesj:
Awesome stuff. I'll try and get around to testing this today.

milesj:
Sorry just been busy. Will have it out today or this weekend.

RobertVillalba:
> Sorry just been busy. Will have it out today or this weekend.

Absolutely understandable thank you for looking into it!

PR #828, opened 2025-07-11
feat: new tools (dbmate, ruff, vale)

risk1996:
## Summary

Added 3 new `proto` plugins, all in TOML format:
 - [`dbmate`](https://github.com/amacneil/dbmate)
 - [`ruff`](https://docs.astral.sh/ruff/) (there is an existing `ruff` plugin by @Phault but cannot be used to install the latest version of `ruff`)
 - [`vale`](https://vale.sh)

### DIFF
--- a/registry/data/third-party.json
+++ b/registry/data/third-party.json
@@ -270,6 +270,19 @@
       ]
     },
     {
+      "id": "dbmate",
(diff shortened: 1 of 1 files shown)


milesj:
Thanks!

Issue #841, opened 2025-07-28
registry improvments


harlequin:
### What version?

0.51.4

### Which command?

proto install TOOL VERSION --log trace

### What happened?

I have tried the implemented `registry` feature today and found some improvements:

1.) Update documentation

https://moonrepo.dev/docs/proto/config#unstable-registries refers to a setting which is not toml compatible

```
[settings]
unstable-registries = [
  { registry: "ghcr.io", namespace: "moonrepo" }
]
```

But it has to be

```
[settings]
unstable-registries = [
  { registry = "ghcr.io", namespace = "moonrepo" }
]
```

Additional the mention of the `default-registry` is missing

2.) Only one registry will be used

In the loader logic `https://github.com/moonrepo/proto/blob/v0.51.4/crates/core/src/loader.rs#L62` only one registry, which is equal to the `default-registry` will be used for searching plugins.

It would be good if the `default-registry` for the `moonrepo` plugins will be used, but if users have own registries with closed source plugins, also these registries will be scanned.

3.) `default-registry` and `unstable-registry` type mismatch

`default-registry` is implemented as string and registry as a type of `registry` and `namespace`.

The user have to set now the following settings to getting stuff working.

```
[settings]
auto-install = true
default-registry = "ghcr.io/moonrepo"
unstable-registries = [
  { registry = "ghcr.io/moonrepo" }
]
```

We have two options here
- Either, the default-registry will be the same type as `unstable-registries`
- Or, `unstable-registries` is getting String values

Both will work and will lead into a consistent behaviour.

**Side Note**:
I have tested also the `registry://` locator in `moon` for toolchain plugins, and it's directly working out of the box!
Perfect.
Next improvement for moon would be the loader logic (incl. settings, like `default-registry` and `registries`)

### Operating system?

Linux

### Architecture?

x64

milesj:
Changed how this works in v0.57.

PR #843, opened 2025-07-29
fix: Scan first default and later all registries

harlequin:
Logic has revisited and the `default-registry` is kept as simple String.
Because: `default-registry` can be modified via env variable, where a String is much easier to handle.

Additional the loader logic for registries has been adapted to a 2 step solution

1) try to get the `first` entry in `registries` which is equal to the `default-registry`

    With this we can define the `default-registry` as `ghcr.io/moonrepo` and the `registry` vector as `{ registry = "ghcr.io/moonrepo" }`

    Yes, it's possible to get the plugins from a registry without defining the `namespace` separately ;)

2) If stage (1) doesn't found anything, it will scan through all defined registries

    With this stage we are enabling the user to define `custom` registries which will taken into consideration without setting the `default-registry` to an exact point. This brings more flexibility for use cases like, `python` is stored in the `default-registry`, `matlab` in another registry and `cmake` again in another registry.

Partially solves #841

### DIFF
--- a/crates/core/src/loader.rs
+++ b/crates/core/src/loader.rs
@@ -58,24 +58,44 @@ pub fn locate_tool(id: &Id, proto: &ProtoEnvironment) -> Result<PluginLocator, P
         }
     }
 
-    // Search in registries
-    if locator.is_none()
-        && let Some(default_registry) = &config.settings.default_registry
-        && let Some(registry) = config
-            .settings
-            .registries
-            .iter()
-            .find(|reg| &reg.registry == default_registry)
-    {
-        if let Ok(maybe_locator) =
-            PluginLocator::try_from(format!("registry://{}", registry.get_reference(id)))
-        {
-            debug!(
-                plugin = maybe_locator.to_string(),
-                "Using a registry plugin"
-            );
+    // Search in registries which fits to the default registry
+    if locator.is_none() {
+        // Search first in default registries, can be multiple namespaces, first wins
+        if let Some(default_registry) = &config.settings.default_registry {
+            debug!(default_registry, "Search first default registry in config");
 
-            locator = Some(maybe_locator.to_owned());
+            locator = config
+                .settings
+                .registries
+                .iter()
+                .filter(|&req| req.registry == *default_registry)
+                .filter_map(|reg| {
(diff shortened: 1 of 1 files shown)


milesj:
The main reason I haven't completed the default registry is because it doesn't work for stuff like this, where the same registry host is used multiple times.

```toml
[settings]
default-registry = "ghcr.io" # which one???
unstable-registries = [
  { registry = "ghcr.io", namespace = "foo" },
  { registry = "ghcr.io", namespace = "bar" },
  { registry = "ghcr.io", namespace = "baz" },
]
```

I was then thinking about adding a `default` property to each registry, but this doesn't work exactly when merging multiple configs.

```toml
[settings]
unstable-registries = [
  { registry = "ghcr.io", namespace = "foo" },
  { registry = "ghcr.io", namespace = "bar", default = true },
  { registry = "ghcr.io", namespace = "baz" },
]
```

Only other idea I have is an `order` property. But unsure if that's good either.

```toml
[settings]
unstable-registries = [
  { registry = "ghcr.io", namespace = "foo" },
  { registry = "ghcr.io", namespace = "bar", order = 1 },
  { registry = "ghcr.io", namespace = "baz", order = 2 },
]
```

harlequin:
Can WE Not imply that the order in the toml file is the Order of vector.

Means in Case of multiple... first wins

milesj:
The order is from lowest depth to highest depth, so `~/.proto/.prototools`, then `~/.prototools`, then children, etc.

It gets a bit more confusing when using env mode files, like `.prototools.production`, or `--config-mode`. I can't expect most users to really understand it. They just expect things to "work".

If we resolved based on that order, `~/.proto/.prototools` would always win (which is where proto would inject moonrepo registries), so how does a user override that?

harlequin:
You are absolute right...grrr

harlequin:
The only problem right would be that the unstable-registry settings will not Work, because the user isn't aware of the default-registry. Which is mandatory to get an entry as maybe_locator

harlequin:
@milesj: I have removed the `default-registry` logic in loader and still kept the for each logic.
As we have discussed.

milesj:
@harlequin Went ahead and made the changers in 0.51.5, give that a try.

PR #844, opened 2025-07-31
refactor: Use enum to unify function string literals.

Binlogo:
…and add documentation for convenience.

## Problem You Are Trying to Solve

There are many plugin capabilities accessed via `tool.plugin.has_func("xxx")`, but we are hardcoding those function names in various places. This makes it challenging for plugin developers to identify them.

## Description

- Introduce enumeration of all available plugin functions that can be implemented by plugins.
- Enhanced documentation for all plugin function with its associated input/output types.
- Updated hardcoded strings with predefined `PluginFunction` and `HookFunction`

## Screenshot

image

### DIFF
--- a/crates/cli/src/commands/install.rs
+++ b/crates/cli/src/commands/install.rs
@@ -7,7 +7,7 @@ use crate::workflows::{InstallOutcome, InstallWorkflowManager, InstallWorkflowPa
 use clap::Args;
 use iocraft::prelude::element;
 use proto_core::{ConfigMode, Id, PinLocation, Tool, ToolSpec};
-use proto_pdk_api::InstallStrategy;
+use proto_pdk_api::{InstallStrategy, PluginFunction};
 use starbase::AppResult;
 use starbase_console::ui::*;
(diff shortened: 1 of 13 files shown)


milesj:
I'm not a fan of this approach. These names will never change, so theres no need for a constant.

If anything, an enum would be far more preferable.

Binlogo:
> I'm not a fan of this approach. These names will never change, so theres no need for a constant.
> 
> If anything, an enum would be far more preferable.

Thanks for the suggestion! Using an enum is definitely more idiomatic in Rust. I’ll update the code to use an enum.

Binlogo:
> > I'm not a fan of this approach. These names will never change, so theres no need for a constant.
> > If anything, an enum would be far more preferable.
> 
> Thanks for the suggestion! Using an enum is definitely more idiomatic in Rust. I’ll update the code to use an enum.

@milesj Hi Miles, I've updated the code to use enums to represent the plugin functions. Please take a look and let me know if there are any other issues. Thanks.

milesj:
Awesome, this is much cleaner, thanks!

PR #903, opened 2025-11-26
new: OpenJDK and Gradle plugins

eplightning:
Adds new entries for OpenJDK Eclipse Adoptium distribution and Gradle plugins.

OpenJDK is a WASM plugin because it needs to use API to discover versions and URLs and deal with various OpenJDK versioning oddities.

### DIFF
--- a/registry/data/third-party.json
+++ b/registry/data/third-party.json
@@ -508,6 +508,19 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


milesj:
Thanks! Theses errors are unrelated to your change.

PR #916, opened 2026-01-08
Add third-party plugin for `geoipupdate`

agdphd:
`geoipupdate` is a tool to manage GeoIP2 and GeoLite database caches, which are used to geolocate IP addresses. I added a Proto plugin definition in https://github.com/maxmind/geoipupdate/pull/453, so I'd like to add it to the public tool registry as well.

### DIFF
--- a/registry/data/third-party.json
+++ b/registry/data/third-party.json
@@ -442,6 +442,19 @@
       ]
     },
     {
+      "id": "geoipupdate",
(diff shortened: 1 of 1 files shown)


milesj:
Cool thanks!

jpsilva:
This broke the registry. There's a comma missing on line 449.

PR #935, opened 2026-01-27
Add automatic bin-to-tool resolution in proto run

### LINKED ISSUE #930: Proto shims seem to prevent npm OIDC environment variables from reaching npm publish
Hey! I've been trying to migrate our monorepo to npm's new trusted publishing (OIDC) feature, and I've run into what seems like an issue with proto's shim mechanism. I'm curious if anyone else has hit this or if there's a workaround I'm missing.

## What I'm trying to do

We use proto (via `moonrepo/setup-toolchain@v0`) in GitHub Actions to manage our Node.js/npm/pnpm versions. I'm trying to use npm's OIDC-based trusted publishing, which requires npm 11.5.1+ and relies on GitHub Actions setting these environment variables:
- `ACTIONS_ID_TOKEN_REQUEST_URL`
- `ACTIONS_ID_TOKEN_REQUEST_TOKEN`

The workflow has `id-token: write` permission, and we've configured trusted publishers on npmjs.com.

## What's happening

When I run `npm publish` (or `pnpm publish` which calls npm), I get `ENEEDAUTH` errors even though everything else is configured correctly. npm isn't seeing the OIDC environment variables it needs.

## Testing I did

I did a bunch of testing to narrow this down:

1. **Verified the env vars are present in the shell** - Added a step in the workflow to echo these variables, and they're definitely there in the GitHub Actions environment.

2. **Tried explicitly passing env vars through the publish command** - Even when using Node.js `exec()` with explicit `env` options, npm still didn't get the variables.

3. **Bypassed proto completely** - This is where it got interesting. When I changed the workflow to:
   - Use `actions/setup-node` directly
   - Install npm 11.6.2 with `npm install -g npm@11.6.2`
   - Install pnpm with `npm install -g pnpm@10.17.1`
   - Run the build and publish without proto shims
   
   Everything worked perfectly! OIDC authentication succeeded and packages published.

## My setup

- proto version: via `.prototools` (moon = "1.39.1", npm = "11.6.2")
- npm version: 11.6.2
- Node.js version: 20.17.0
- OS: Ubuntu (GitHub Actions runner)
- GitHub Actions with `id-token: write` permission

## What I think is happening

It seems like when proto shims intercept the `pnpm`/`npm` calls, the environment variables don't get inherited all the way through the chain. I'm guessing the shim → `proto run` → actual binary process chain somehow loses the OIDC variables along the way, even though other env vars seem to make it through fine.

## Questions

- Has anyone successfully used npm OIDC publishing with proto shims?
- Is there a way to configure proto to ensure these specific environment variables get passed through?
- Should I be using a different approach when publishing from CI/CD with proto?

For now, I've worked around it by not using proto in our release workflow, but I'd love to use proto consistently across all our CI jobs if there's a way to make this work. Any ideas?

---

**Relevant links:**
- npm OIDC documentation: https://docs.npmjs.com/trusted-publishers
- Our testing PRs: 
  - Verification that env vars are present: https://github.com/adobe/spectrum-design-data/pull/687
  - Attempt with explicit env passing: https://github.com/adobe/spectrum-design-data/pull/688
  - Successful publish bypassing proto: https://github.com/adobe/spectrum-design-data/pull/690

GarthDB:
Fixes #930 - Adds automatic bin-to-tool resolution for commands like `npx`, `bunx`, etc.

When running `proto run npx`, proto now:
1. Detects that "npx" is not a registered tool
2. Queries the plugin registry to find which tool provides the "npx" bin
3. Discovers that "npm" provides "npx"
4. Automatically redirects to use proto's managed npm with npx as the executable

This allows natural usage of `proto run npx` instead of requiring `proto run npm -- exec`.

## Changes

- **Modified `crates/cli/src/commands/run.rs`**: Added bin-to-tool lookup on UnknownTool error
- **Added comprehensive tests**: 6 new tests covering npx, bunx, node-gyp, passthrough args, and fallback
- **Updated `CHANGELOG.md`**: Documented the new feature

## Testing

- All new tests pass (6 passed, 0 failed)
- Validated with real-world repos (changesets publishing workflows)
- Tested with Node 18-24, npm 9-11, bun 1.2-1.3

## Backward Compatibility

- No breaking changes
- Fallback behavior preserved
- Existing tools and configurations unaffected

### DIFF
--- a/crates/cli/src/commands/run.rs
+++ b/crates/cli/src/commands/run.rs
@@ -265,6 +265,8 @@ pub async fn run(session: ProtoSession, mut args: RunArgs) -> AppResult {
 
             let registry_path = session.env.store.shims_dir.join("registry.json");
             let mut parent_tool_id: Option<Id> = None;
+            let mut before_args: Vec<String> = vec![];
+            let mut after_args: Vec<String> = vec![];
 
             // Try reading the shims registry
             if registry_path.exists() {
@@ -282,6 +284,10 @@ pub async fn run(session: ProtoSession, mut args: RunArgs) -> AppResult {
                                             parent
                                         );
                                         parent_tool_id = Some(Id::raw(parent));
+
+                                        // Store before/after args from the shim entry
+                                        before_args = shim_entry.before_args.clone();
+                                        after_args = shim_entry.after_args.clone();
                                     }
                                 }
                             }
@@ -311,6 +317,12 @@ pub async fn run(session: ProtoSession, mut args: RunArgs) -> AppResult {
                 // Update args to run the parent tool with this bin as an alternate executable
                 args.exe = Some(id.to_string());
                 args.context = ToolContext::new(parent_id);
+
+                // Prepend before_args and append after_args to passthrough
+                let mut new_passthrough = before_args;
+                new_passthrough.extend(args.passthrough.clone());
+                new_passthrough.extend(after_args);
+                args.passthrough = new_passthrough;
 
                 // Load the parent tool (this will handle auto-install if enabled,
                 // or show a proper error message if the tool is not installed)


GarthDB:
@milesj I successfully ran the GitHub Actions locally with [`act`](https://github.com/nektos/act), so I think it's ready now.

crates/cli/tests/run_test.rs
```diff
@@ -611,4 +611,312 @@ FOURTH = "ignores-$FIRST-$PARENT"
...
+                .run_bin(|cmd| {
+                    cmd.args(["install", "npm", "9.0.0", "--pin"]);
+                })
+                .success();
+
+            // Verify npx can be run directly (should redirect to npm)
+            // Just check it executes without error
+            sandbox
+                .run_bin(|cmd| {
+                    cmd.args(["run", "npx", "--help"]);
```
milesj: @GarthDB I think this (and some others below) need to be `-- --help` instead of `--help`, as I believe the latter is showing `proto run`s help menu, not the tool's help menu.

GarthDB: Good catch, I'll adjust it.

GarthDB: @milesj Good catch on the `before_args` and `after_args`. I think I've updated the implementation to handle these properly now.

When resolving a bin from the shims registry (lines 266-325), it now:
1. Extracts `before_args` and `after_args` from the shim entry
2. Reconstruct the passthrough arguments as: `before_args` + `user_args` + `after_args`

I think this would ensure that alternate binaries like `bunx` (which needs `["x"]` prepended to become `bun x`) work correctly. All tests pass locally with this change.

Let me know if this looks good or if there's a better approach.

milesj:
Awesome  stuff!

PR #979, opened 2026-04-05
fix: avoid redundant shim permission updates during lookup

rotu:
## Summary
Skip permission updates during executable lookup when the target file is already executable.

## Why
In the VS Code Copilot sandbox, npm shims under ~/.proto/tools/npm/.../shims fail with EPERM when proto re-runs chmod on every lookup, even when the shim already has executable permissions.

## Changes
- add a small helper in locate flow to decide whether permissions actually need updating
- only call update_perms when the target exists and is not already executable

## Validation
- built local proto/proto-shim successfully
- reproduced the original sandbox failure before the change
- verified the patched local build succeeds for `proto run npm -- --version` in the sandbox
- verified the patched local proto-shim path succeeds in the sandbox when pointed at the local debug binary

### DIFF
--- a/crates/core/src/flow/locate.rs
+++ b/crates/core/src/flow/locate.rs
@@ -1,6 +1,6 @@
 pub use super::locate_error::ProtoLocateError;
-use crate::helpers::is_executable;
 use crate::helpers::ENV_VAR;
+use crate::helpers::is_executable;
 use crate::layout::BinManager;
 use crate::tool::Tool;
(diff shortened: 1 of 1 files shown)


rotu:
PR created with AI. Note it would probably be cleaner to fix the permission issue on tool install, but that would be a more involved PR.

milesj:
Seems reasonable, thanks!

PR #994, opened 2026-04-27
fix: Normalize proto activate paths for emulated POSIX shells on Windows

ourchitectureio:
## Summary

When running `proto activate` for Bash/Zsh/Fish/Murex/Elvish under a
POSIX-emulating shell on Windows (Git Bash, MSYS2, Cygwin), the exported
`PATH` and `_PROTO_ACTIVATED_PATH` previously contained native
`C:\Users\…\proto\bin` entries joined with `;`. Those shells parse `:` as
the separator and treat backslashes as escape characters, so activation
silently produced an unusable `PATH` and proto-managed tools fell off
resolution.

## Fix

Detect emulated POSIX shells on Windows via `cfg!(windows)` plus
`MSYSTEM` / `MINGW` / `MSYS` / `OSTYPE=msys|cygwin`, and translate paths
to POSIX form (`C:\Users\Alice` → `/c/Users/Alice`,
`\\server\share` → `/unc/server/share`) before they are written into
`_PROTO_ACTIVATED_PATH` or emitted via `format_path_set`. The same
helper is also used when producing `--json` output for these shells.

Linux and macOS are untouched (the helper is `cfg(not(windows))` no-op),
and native Windows shells (PowerShell / pwsh / cmd) continue to receive
their existing path format. Command execution (`apply_to_command`) is
unchanged.

## Changes

- `crates/cli/src/workflows/exec_workflow.rs`: shell-aware path joining
  and conversion (`join_paths_for_shell`, `convert_paths_for_shell`,
  `windows_path_to_posix`, `is_windows_posix_shell`); the boolean is
  computed once per call, not per path.
- `crates/cli/src/commands/activate.rs`: routes activation exports and
  the JSON `path` field through the new helpers.
- `crates/cli/tests/activate_windows_test.rs`: Windows-only integration
  tests asserting that bash output uses `/c/...:` separators while pwsh
  output keeps `;` and native paths.
- Unit tests for drive-letter, UNC, already-POSIX, and relative-path
  cases, plus `is_windows_posix_shell` detection. Test env mutations
  use a scoped `unsafe` `EnvVarGuard` to satisfy Rust 2024 and restore
  the original value on panic.
- `CHANGELOG.md`: entry under `## Unreleased` → `#### 🐞 Fixes`.

## Validation

- `just lint` and `just test` (Linux/macOS): existing snapshots
  unchanged, since the new code paths are gated on `cfg!(windows)`.
- `cargo test -p proto_cli --bin proto workflows::exec_workflow::tests`
  on Windows: 11 passed (drive-letter, UNC, already-POSIX, relative,
  shell detection, plus the pre-existing path tests).
- Manual `proto activate bash --export` inside Git Bash (MSYSTEM=MINGW64):
  `_PROTO_ACTIVATED_PATH` and `PATH` now contain `/c/Users/.../shims:/c/Users/.../bin` style entries.
- Manual `proto activate pwsh --export` from the same Windows host: output unchanged from current behavior.

### DIFF
--- a/crates/cli/src/workflows/exec_workflow.rs
+++ b/crates/cli/src/workflows/exec_workflow.rs
@@ -15,6 +15,7 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use starbase_args::parse as parse_args;
 use starbase_shell::{BoxedShell, ShellType, join_args};
 use starbase_utils::envx;
+use std::borrow::Cow;
 use std::collections::VecDeque;
 use std::env;
 use std::ffi::{OsStr, OsString};
@@ -265,6 +266,17 @@ impl<'app> ExecWorkflow<'app> {
         Ok(None)
     }
 
+    pub fn activation_path_value_for_shell(
+        &self,
+        shell_type: &ShellType,
+    ) -> miette::Result<OsString> {
+        join_paths_for_shell(self.paths.iter(), shell_type)
+    }
+
+    pub fn reset_paths_for_shell(&self, store_dir: &Path, shell_type: &ShellType) -> Vec<String> {
+        convert_paths_for_shell(self.reset_paths(store_dir).iter(), shell_type)
+    }
+
     pub fn reset_paths(&self, store_dir: &Path) -> Vec<PathBuf> {
         let start_path = store_dir.join("activate-start");
         let stop_path = store_dir.join("activate-stop");
@@ -300,8 +312,12 @@ impl<'app> ExecWorkflow<'app> {
         reset_paths
     }
 
-    pub fn reset_and_join_paths(&self, store_dir: &Path) -> miette::Result<OsString> {
-        env::join_paths(self.reset_paths(store_dir)).into_diagnostic()
+    pub fn reset_and_join_paths_for_shell(
+        &self,
+        store_dir: &Path,
+        shell_type: &ShellType,
+    ) -> miette::Result<OsString> {
+        join_paths_for_shell(self.reset_paths(store_dir).iter(), shell_type)
     }
 
     pub fn requires_shell(&self, shell: &BoxedShell, args: &[String], raw: bool) -> bool {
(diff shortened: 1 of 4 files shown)


ourchitectureio:
CI is currently failing on what looks like an unrelated timeout in the test suite: `proto_cli::clean_test::clean::cleans_multiple_stale_tool_versions` (Job `73268863300`). It times out after ~60s (fails twice) rather than failing an assertion, which suggests flakiness / an environment-dependent hang in the `clean` flow, not necessarily an issue introduced by the changes in this PR.

If helpful, I can open one or more follow-up issues to track hardening this area. A few concrete ideas:
- Make the `clean` tests fully hermetic by forcing a temp `PROTO_HOME` / isolated test directories (avoid depending on runner state).
- Add a bounded timeout around the spawned CLI process in the test so it fails fast with clearer diagnostics instead of hanging.
- Ensure `proto clean` is strictly filesystem-only during tests (no network/tool resolution paths), and add timeouts to any lock acquisition if applicable.

Happy to help with a repro or a small targeted follow-up PR once the desired approach is agreed.

crates/cli/src/workflows/exec_workflow.rs
```diff
@@ -325,6 +341,105 @@ impl<'app> ExecWorkflow<'app> {
...
+        || env::var("OSTYPE")
+            .map(|value| {
+                let value = value.to_ascii_lowercase();
+                value.contains("msys") || value.contains("cygwin")
+            })
+            .unwrap_or(false))
+}
+
+#[cfg(windows)]
+fn windows_path_to_posix(path: &Path) -> Cow<'_, str> {
```
milesj, line 408: Looping over the path components would probably be easier here, and more reliant.

crates/cli/src/workflows/exec_workflow.rs
```diff
@@ -563,5 +678,72 @@ mod tests {
...
+                "relative\\bin"
+            );
+        }
+
+        #[cfg(windows)]
+        #[test]
+        fn detects_emulated_posix_shells() {
+            use std::env;
+
+            struct EnvVarGuard {
```
milesj: This guard isn't necessary. Just call the env funs directly.

milesj (changes requested):
@ourchitectureio Just a few minor comments but looks good otherwise.

crates/cli/src/workflows/exec_workflow.rs
milesj: Also found this, which may work too: https://crates.io/crates/cross-path

milesj:
@ourchitectureio You able to finish this?

crates/cli/src/workflows/exec_workflow.rs
ourchitectureio: Yes. I removed the guard approach. I ended up going one step further and refactored the detection helper so the test can inject env values directly, which avoids mutating process-global environment state during the test.

crates/cli/src/workflows/exec_workflow.rs
ourchitectureio: Thanks, agreed. I updated the conversion to iterate over `std::path::Component` / `Prefix` instead of doing manual string parsing. That ended up being simpler and more robust for Windows drive and UNC path handling.

Thanks for the pointer on "cross-path". I took a similar direction conceptually, but kept it dependency-free since the conversion we need is fairly narrow and is now handled cleanly with `std::path` components.

ourchitectureio:
Thanks for the review and for the patience here @milesj and team. I’ve pushed follow-up changes addressing the requested updates: the Windows path conversion now uses path components instead of manual string handling, and the shell-detection test was simplified by removing the guard approach and avoiding process-global env mutation entirely. I also reran the targeted build and relevant Windows activation tests locally, and they passed.

ourchitectureio:
I've removed the CHANGELOG entry to avoid any assumptions about the change process as well as merge conflicts.

Please feel free to edit the PR and code if a CHANGELOG update is recommended as you are likely in the best position to ensure the right updates.

milesj:
@ourchitectureio Thanks looks good! I'll handle the formatting issue.

Issue #1006, opened 2026-05-11
Node 24 binary missing on Windows after install (proto::locate::missing_executable)


rhuanbarreto:
## Environment

- **proto version**: 0.56.4
- **OS**: Windows Server 2025 (GitHub Actions `windows-latest`)
- **Node version requested**: `~24` (resolved to 24.14.0)
- **.prototools**: `node = "24"`

## Problem

When installing Node 24 on Windows via `moonrepo/setup-toolchain`, proto reports the installation as successful but the actual `node.exe` binary is never written to disk. The warnings are emitted during symlink creation but the install is not marked as failed:

```
[ WARN ] proto_core::flow::link  Unable to symlink binary, source file does not exist
  tool="node"
  source="C:\Users\runneradmin\.proto\tools\node\24.14.0\node.exe"
  target="C:\Users\runneradmin\.proto\bin\node.exe"

[node] Node.js 24.14.0 installed
```

After this, any attempt to use `node` via the proto shim fails:

```
Error: proto::locate::missing_executable

  × Unable to find an executable for Node.js, expected file
  │ ~\.proto\tools\node\24.14.0\node.exe does not exist.
```

## Reproduction

1. Create `.prototools` with `node = "24"`
2. Run `moonrepo/setup-toolchain` on `windows-latest` with `auto-install: true`
3. Try to run `node --version`

## Expected behavior

Either:
- The Node 24 binary should be properly downloaded and installed on Windows
- Or the installation should fail with an error (not succeed with warnings) so the CI step fails early with a clear message

## Workaround

Pinning to `node = "22"` (LTS) works correctly on Windows.

## CI log reference

Full log: https://github.com/archgate/cli/actions/runs/25657982746/job/75311030982

milesj:
@rhuanbarreto This is confusing since I've been testing with v24 non-stop for the past week on this PR: https://github.com/moonrepo/proto/pull/996

Works just fine on all operating systems. 

Can you `ls` the directory to see what's unpacked?

rhuanbarreto:
@milesj Thanks for looking into this. You're right — `proto install node 24` works fine locally on my Windows machine too (proto 0.56.4, Windows 10 x86_64, Git Bash). The issue was specific to **GitHub Actions `windows-latest` runners using `moonrepo/setup-toolchain`**.

Here's what we observed in the CI logs:

**Run 1 (with cache hit):**
```
Cache hit for: moonrepo-toolchain-v3-win32-x64-368f3382...
[...]

PR #1007, opened 2026-05-11
fix: verify primary binary exists after installation

rhuanbarreto:
## Summary

After extracting a tool archive, proto now checks that the primary executable actually exists on disk before reporting the install as successful. This prevents a class of bugs where a silently failed extraction is reported as "installed", only to fail later in the link phase with a confusing warning.

## Problem

On Windows with proto 0.56.4, installing Node 24 (via `setup-toolchain` in GitHub Actions) produces:

```
[node] Node.js 24.14.0 installed

[ WARN ] proto_core::flow::link  Unable to symlink binary, source file does not exist
  tool="node" source="~\.proto\tools\node\24.14.0\node.exe"
```

The install reports success, but `node.exe` was never written to disk. Downstream consumers (postinstall scripts, shims) fail with `proto::locate::missing_executable` and no clear indication of what went wrong.

This was originally reported in #1006.

## Fix

Added a post-install verification step in `Installer::install()` (crates/core/src/flow/install.rs) that:

1. Calls `locate_executables` to get the expected primary binary path
2. Checks if that file exists on disk
3. If missing, returns a new `MissingBinaryAfterInstall` error with an actionable message:

```
Installation of Node.js completed but the expected binary
~/.proto/tools/node/24.14.0/node.exe does not exist. The archive
may have failed to extract correctly, or the download may be corrupt.
Try running `proto install --force` to re-install.
```

## Changes

- `crates/core/src/flow/install.rs` — added binary existence check after archive extraction and post-install scripts
- `crates/core/src/flow/install_error.rs` — added `MissingBinaryAfterInstall` error variant

## Test plan

- `cargo check --package proto_core` passes
- The check uses the same `locate_executables` plugin call that the linker uses, so the expected path is guaranteed to match what the link phase would look for

### DIFF
--- a/crates/cli/tests/install_one_test.rs
+++ b/crates/cli/tests/install_one_test.rs
@@ -829,13 +829,13 @@ mod install_one {
     fn errors_when_primary_binary_missing_after_install() {
         let sandbox = create_empty_proto_sandbox();
 
-        // Version 0.0.1 triggers the mocked tool's "broken install" mode:
+        // Version 1.0.1 triggers the mocked tool's "broken install" mode:
         // native_install creates secondary executables but skips the primary
         // binary, simulating a failed archive extraction where the binary
         // was never written to disk.
         let assert = sandbox
             .run_bin(|cmd| {
-                cmd.arg("install").arg("protostar").arg("0.0.1");
+                cmd.arg("install").arg("protostar").arg("1.0.1");
             })
             .failure();
 
--- a/crates/cli/tests/install_one_test.rs
+++ b/crates/cli/tests/install_one_test.rs
@@ -825,6 +825,23 @@ mod install_one {
         }
     }
 
+    #[test]
+    fn errors_when_primary_binary_missing_after_install() {
+        let sandbox = create_empty_proto_sandbox();
+
+        // Version 0.0.1 triggers the mocked tool's "broken install" mode:
+        // native_install creates secondary executables but skips the primary
+        // binary, simulating a failed archive extraction where the binary
+        // was never written to disk.
+        let assert = sandbox
+            .run_bin(|cmd| {
+                cmd.arg("install").arg("protostar").arg("0.0.1");
+            })
+            .failure();
+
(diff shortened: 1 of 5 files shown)


rhuanbarreto:
## CI Test Failures — Triage

**My changes (binary verification):** The new `errors_when_primary_binary_missing_after_install` test and the binary check code should work correctly after the latest push (version `1.0.1` trigger, `no_bin` skip, `verify_primary_binary` extraction).

**Pre-existing failures unrelated to this PR:**

1. **`clean_test::cleans_multiple_stale_tool_versions`** — Flaky, fails on `master` too (verified locally). Timing/network-dependent.

2. **`plugins_test::supports_node_and_package_managers`** — pnpm 11.0.9 (latest) now requires Node ≥ 22.13, but the test sandbox installs Node 20.x. The test runs `proto run pnpm -- --version` which crashes with `ERR_UNKNOWN_BUILTIN_MODULE: No such built-in module: node:sqlite`. This broke when pnpm shipped v11 with the Node 22.13 minimum requirement — it's a version pinning issue in the test, not related to this PR. The snapshot update (adding `pn`/`pnx`) is correct for pnpm 11.

Both failures reproduce on `master` — they're not regressions from this PR.

milesj:
@rhuanbarreto There's already a method for this, `locate_exe_file`.

rhuanbarreto:
@milesj Good call — refactored to use `Locator::locate_exe_file()` instead of the custom implementation. Much cleaner: removed the custom `MissingBinaryAfterInstall` error variant and `verify_primary_binary` method in favor of a single `Locator` call that reuses the existing `MissingToolExecutable` error. See latest commit.

Also noting that the pre-existing test failures (`supports_node_and_package_managers` + `clean_test` flakes) are already addressed in your #1001 branch, so those aren't blockers for this PR.

For disclosure: this is Claude Opus 4.6 writing the comment. But I'm the ultimate responsible and I'm reviewing everything.

PR #1020, opened 2026-06-02
Update OCaml plugin locator

Hebilicious:
Updates the OCaml third-party plugin entry to point at the new Hebilicious/proto-plugins monorepo locator.


Hebilicious:
Superseded by #1022, which updates both Hebilicious plugin locators together.

PR #1021, opened 2026-06-02
Add Nushell plugin

Hebilicious:
Adds the Nushell third-party WASM plugin entry and points it at the new Hebilicious/proto-plugins monorepo locator.


Hebilicious:
Superseded by #1022, which uses a fresh upstream-based branch and updates both Hebilicious plugin locators together.

PR #1030, opened 2026-06-15
feat: allow overriding default registry host and namespace via env vars

harlequin:
`RegistryConfig.default` is now Option<bool> so missing values can be distinguished from false locate_plugin now treats absent default as true, preserving the old behavior for legacy configs

.This avoids breaking old config files that did not specify default, while still allowing new configs to explicitly disable the default registry with default: `false`

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -666,8 +666,22 @@ pub struct ProtoConfigEnvOptions<'ctx> {
 
 pub fn find_debug_locator_with_fallback(name: &str, version: &str) -> PluginLocator {
     static URL_CACHE: OnceLock<bool> = OnceLock::new();
+    static MOON_REGISTRY_HOST_CACHE: OnceLock<String> = OnceLock::new();
+    static MOON_REGISTRY_NAMESPACE_CACHE: OnceLock<String> = OnceLock::new();
 
     let use_urls = *URL_CACHE.get_or_init(|| bool_var("PROTO_PLUGINS_USE_URL_DIST"));
+    let moon_registry_host = MOON_REGISTRY_HOST_CACHE.get_or_init(|| {
+        env::var("MOON_DEFAULT_REGISTRY_HOST")
+            .ok()
+            .filter(|s| !s.is_empty())
+            .unwrap_or_else(|| "ghcr.io".to_string())
+    });
+    let moon_registry_namespace = MOON_REGISTRY_NAMESPACE_CACHE.get_or_init(|| {
+        env::var("MOON_DEFAULT_REGISTRY_NAMESPACE")
+            .ok()
+            .filter(|s| !s.is_empty())
+            .unwrap_or_else(|| "moonrepo".to_string())
+    });
 
     find_debug_locator(name).unwrap_or_else(|| {
         if use_urls {
@@ -678,8 +692,8 @@ pub fn find_debug_locator_with_fallback(name: &str, version: &str) -> PluginLoca
             }))
         } else {
             PluginLocator::Registry(Box::new(RegistryLocator {
-                registry: Some("ghcr.io".into()),
-                namespace: Some("moonrepo".into()),
+                registry: Some(moon_registry_host.into()),
+                namespace: Some(moon_registry_namespace.into()),
                 image: name.into(),
                 tag: Some(version.into()),
             }))
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -1,4 +1,5 @@
 use crate::config_error::ProtoConfigError;
+use crate::get_builtin_registry;
 use crate::helpers::{ENV_VAR_SUB, fast_map_clone};
 use crate::tool_context::ToolContext;
 use crate::tool_spec::ToolSpec;
@@ -666,22 +667,9 @@ pub struct ProtoConfigEnvOptions<'ctx> {
 
 pub fn find_debug_locator_with_fallback(name: &str, version: &str) -> PluginLocator {
     static URL_CACHE: OnceLock<bool> = OnceLock::new();
-    static MOON_REGISTRY_HOST_CACHE: OnceLock<String> = OnceLock::new();
-    static MOON_REGISTRY_NAMESPACE_CACHE: OnceLock<String> = OnceLock::new();
 
     let use_urls = *URL_CACHE.get_or_init(|| bool_var("PROTO_PLUGINS_USE_URL_DIST"));
-    let moon_registry_host = MOON_REGISTRY_HOST_CACHE.get_or_init(|| {
-        env::var("MOON_DEFAULT_REGISTRY_HOST")
-            .ok()
-            .filter(|s| !s.is_empty())
-            .unwrap_or_else(|| "ghcr.io".to_string())
-    });
-    let moon_registry_namespace = MOON_REGISTRY_NAMESPACE_CACHE.get_or_init(|| {
-        env::var("MOON_DEFAULT_REGISTRY_NAMESPACE")
-            .ok()
-            .filter(|s| !s.is_empty())
-            .unwrap_or_else(|| "moonrepo".to_string())
-    });
+    let builtin_registry = get_builtin_registry().to_owned();
 
     find_debug_locator(name).unwrap_or_else(|| {
         if use_urls {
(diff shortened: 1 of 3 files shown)


milesj:
@harlequin Well, these were breaking changes on purpose :P

harlequin:
But why? 

With this merge request its possible to download the official plugins from ghrcio and additional some closed source from our internal repo. 

Is there any roadmap Item which i have missed

milesj:
An undefined field magically becoming the default feels like a footgun. Ideally only 1 registry is marked as default, any others should not be.

harlequin:
In our company we have at least 2 divisions with an own repo, plus your official one. So plugins shall bei detected, with only the short form name, from there. 

Would bei great ti implement something Like first hit wins

harlequin:
@milesj Maybe we have to discuss about the implementation details and the "breaking" changes

You have released several versions with this settings footprint

1. First version

```
[settings]
default = "custom.host.com"
unstable-registries = [
  { registry = "custom.host.com" }
]
```

2. Second version

here the `default` attribute was removed as breaking change and a migration path was performed from our side that all "internal" users have a correct registry setting.

```
[settings]
unstable-registries = [
  { registry = "custom.host.com" }
]
```

3. Current Version

With the current release you have introduced back the `default` property, but now inside the registry config block. This would again lead into a migration path for all our "internal" users and projects.

```
[settings]
unstable-registries = [
  { registry = "custom.host.com", default = true }
]
```




I try now to write down the uses cases for which and how such a registry is useful, escp. on our side

1. Use Case - Normal Operation

  Settings does **not** have any `unstable-registries` set

  This shall download the plugins from ghcr.io

2. Use Case - Remote/Mirror Registry

  Many `registries` have the ability to mirror a registry hosted somewhere, to ensure a `air gapped` environment and even with scanners to avoid `Supply Chain Attacks`

  Settings have **one** entry for unstable-registries`

  This shall download the plugins from custom.host.com which is a mirror from ghcr.io

3. Use Case - Many registries are defined

  This would be the case if a company or even a division within the company has custom plugins for proto and moon

  This shall search and download the plugins from the given registry (first hit wins)
  This would even imply that when a user sets a custom registry he has to set also the official ghcr.io or the `mirror` from use case 2 into the registry config block

  Example for official and custom

  ```
  unstable-registries = [
    { registry = "ghcr.io" }
    { registry = "custom.host.com" }
  ]
  ```

  Example for mirror and custom
  ```
  unstable-registries = [
    { registry = "mirror.host.com" }
    { registry = "custom.host.com" }
  ]
  ```


For use case (3) I think one way of an implementation would be the `default` setting.
When declaring this as `true` it will **replace** the core ghcr.io and to download the plugins from there. --> Mirror functionality

If from the list of registries none are set to `default = true` it will download the official ones from ghcr.io and searches the other registries for custom plugins. This would allow users from division inside a company to deploy and release custom plugins with a custom registry.



What do you think?
Did I missed a use case?

If you are agreeing to it I would directly start to contribute within this pull request to the described behavior.

This registry is very important for us as a company, because it helps alot to distribute closed source plugins and for the end user and easy way for declaring and installing.

milesj:
The discussion around `default` only applies when a tool with an unknown ID is used and no plugin has been defined. For example:

```
foo = "1.2.3"
```

What is `foo`, where does it come from? When there is a registry marked as default, we turn this into `registry://host.com/owner.foo` behind the scenes. That's all that default does.

Both 1 and 2 use cases don't apply here. Even our built-in plugins don't hit this code path. This merely exists for consumers.

All cases only apply for this scenario, where the host/namespace is unknown.

```
foo = "1.2.3"

[plugins.tools]
foo = "registry://foo"
```

In this scenario, we loop through each registry until we get a hit. Otherwise, use the full path: `registry://host.com/owner/foo`.

> If from the list of registries none are set to default = true it will download the official ones from ghcr.io and searches the other registries for custom plugins. 

If no plugin locator is defined, and no registry is marked as default, then its simply a hard error. It works this way because we don't have a central registry. Additionally, searching through all registries for a match is MASSIVE NETWORK hit. proto public users would absolutely not like this, as ghcr.io is already very slow and somewhat unreliable.

Additionally, the moonrepo ghcr.io _is configured_ in `unstable-registries`, and will always appear in the list. There is no way to overwrite.

harlequin:
Hey @milesj, I get your points

but I have two last questions which have to answer me ;)

1. When only the registry with `default` will be taken into consideration than my use case would be

```
unstable-registries = [
  { registry = "mirror-from-ghcr.io", default = true } // to a avoid supply chain attacks and air gapped env
  { registry = "custom-division-1.host.com", default = true } // for in house registry with closed source plugins
  { registry = "custom-division-2.host.com", default = true } // for other house registry with closed source plugins
]
```

Because we want to achieve that users could install, for example rhapsody simply with

`proto install rhapsody`

And what I have understood from the source code `loader.rs` "only" the default ones are iterated and searched




### Alternative

When I have understood everything completely wrong, my fallback solution would be
to create a so called virtual repository within artifactory which internally points 
to all 3 listed registries above and overlay these.
This would result into a prototools config block like this

```
unstable-registries = [
  { registry = "virtual-repo.inhouse.com", default = true }
]
```

@milesj: which way you would suggest?


2. What is exactly missing to get this feature as `stable`

I want to ensure that if there will be a migration path, to do it only one more time ;)

milesj:
> When only the registry with default will be taken into consideration than my use case would be

Only the first default registry found will be used.

----

What it sounds like you want is this: when there is no plugin locator defined for a tool identifier, loop through all registries until one is found?

If so, that doesn't exist currently. What does exist though, is if the identifier has a plugin locator of simply `registry://<id>`, with no host/owner set, then we _do_ loop through all registries till we find a plugin.

So the question becomes, how do we assign an empty registry plugin locator for a tool identifier? `default` only uses the first registry found, not all.

I think we need an entirely new setting for this, something like `on-missing-plugin = 'registry-fallback'` or something.

harlequin:
> What it sounds like you want is this: when there is no plugin locator defined for a tool identifier, loop through all registries until one is found?

Yes this is exactly what I am demanding for.

All package manager which I am aware of are working like this, e.q. apt or node
You can add as many repo sources as you want but first hit wins.

Even when the `search` functionality will be available for registries
It shall also scan for all configured ones.

If end users would have troubles with too many registries and too many iterations, they would have the possibility to specify the plugin location.


I would implement it like this:-

Nothing is configured, so the `default` will be put and proto is working as intended.

A user has configured at least one registry but **without** default, it would take your `default` as first entry in the registry list and add(s) the configured registries into the list.

A user has configured at least one registry **with** default, it would take this as your ghcr replacement (which would also would save bandwidth for your official registry) and add the rest to the list.

In case a user has configured multiple `default` in the settings block, warn user that only the **first** will be taken as replacement/mirror for the ghcr.


But I have investigated the source code and it really hard to implement this, because what you have said, the plugin locator doesn't iterate and have not "easily" a method for check if artifact exists in registry.

Let me think about this all and I will update the merge request (but this will take a while)

milesj:
> All package manager which I am aware of are working like this, e.q. apt or node

All of these managers have a central repository, where anyone can publish to it. One source of truth.

proto does not work that way, there is no central repository, just many isolated repositories. The same concepts don't apply here.

Think of it like monorepo vs polyrepo.

> But I have investigated the source code and it really hard to implement this, because what you have said, the plugin locator doesn't iterate and have not "easily" a method for check if artifact exists in registry.

Yeah this isn't easy, maybe impossible right now. All of the OCI logic is in warpgate, but the plugin locator resolution is in proto. Warpgate is agnostic and not coupled to proto at all, so changing any of this is not ideal.

harlequin:
@milesj I have thought about your concerns and strategies you have pointed out here.

**And you are right, let us keep the architecture of the crates as they are!**

After discussion with my team we have another "wish" which would even help you

- We want to create a mirror of your ghcr.io/moonrepo in our artifactory instance 
- We have already our internal registry in artifactory for the closed source plugins
- We will create a "virtual" repository in artifactory which combines the "remote" and "internal" (just imagine like an fs overlay)


Now what we would have in mind...

(1) When defining the default it will additional override the "ghcr.io" from the proto builtin plugins

(2) Or, as backup, the url-rewrite shall also support the registry location


Either option 1 or 2 would save you a lot of bandwidth in the ghcr.io 


If you would agree, I can distribute within this PR for (1) or (2)

milesj:
@harlequin Would env vars be enough?

harlequin:
Yes absolute

milesj:
Sounds good, feel free to update this PR to add `MOON_DEFAULT_REGISTRY_HOST`, `NAMESPACE`, `AUTH`, etc, that override the moonrepo ones.

harlequin:
@milesj Have implemented the env vars at the location for the internal plugins.

milesj:
@harlequin Actually, maybe we call it `MOON_BUILTIN_` instead of `MOON_DEFAULT_`, as to not confuse it with `default = true` registries. We probably need an auth var too.

You'll also need to update this location too: https://github.com/moonrepo/proto/blob/master/crates/core/src/settings/settings.rs#L57

harlequin:
Updated as demanded

milesj:
Looks good!

PR #1048, opened 2026-07-11
fix: Recurse into cache subdirectories when cleaning

### LINKED ISSUE #1047: `proto clean cache` skips stale files in subdirectories
### What version?

Observed on 0.55.2; source-confirmed unchanged in 0.58.2 at [390ee275](https://github.com/moonrepo/proto/commit/390ee2751315733c80066a3731be87bae50764d0).

### Which command?

`proto clean cache` (the same helper backs the `temp` and `plugins` targets, so nested files under those roots are subject to the same behavior)

### What happened?

`~/.proto/cache` had grown to ~10 GB across `cache/requests` (orphaned since the switch to `requests-v2`) and `cache/requests-v2` (the plugin loader's HTTP request cache). Every clean invocation reports:

```
Clean complete but nothing was removed.
No artifacts were found older than 30 days.
```

Minimal reproduction:

```sh
mkdir -p ~/.proto/cache/repro && touch ~/.proto/cache/repro/stale.json
touch -a -m -t 202601010000 ~/.proto/cache/repro/stale.json
proto clean cache --days 1 --yes
# stale.json remains and the command reports nothing was removed
```

`clean_dir()` in [`crates/cli/src/commands/clean.rs`](https://github.com/moonrepo/proto/blob/390ee2751315733c80066a3731be87bae50764d0/crates/cli/src/commands/clean.rs#L272-L306) iterates `fs::read_dir(dir)` and only acts on entries where `path.is_file()`. Everything directly under `~/.proto/cache` is a directory (`requests-v2`, `registry`, and the orphaned `requests`), so every entry is skipped before any staleness check runs.

### Operating system?

macOS (arm64)

I have a fix ready with tests and will open a PR linked to this issue.

liby:
Fixes #1047

`clean_dir()` only iterated immediate children and only processed regular files, so `proto clean cache` never removed anything under `~/.proto/cache`: every entry there is a directory (`requests-v2`, `registry`), and the HTTP request cache inside `requests-v2` has no eviction of its own, so it grows without bound.

`clean_dir` now takes a `recurse` flag. For the cache target it recurses into subdirectories, removes stale files at any depth, and prunes a subdirectory once it becomes empty (tolerating `NotFound`/`DirectoryNotEmpty` from concurrent writers). The cache root itself is never removed and dot-prefixed entries keep being skipped. Entries are classified via `DirEntry::file_type()`, so symlinks are neither followed nor deleted.

The other two targets keep their previous top-level behavior deliberately:

- **plugins** stays shallow because the plugin cache layout is flat (`create_cache_path` writes `<plugins>/<id>-<hash>.<ext>`); there is nothing nested to clean.
- **temp** stays shallow because it is the one store directory with concurrent writers: installs hold a directory lock on `<temp>/<tool>/<hash>` (`fs::lock_directory`) and plugin downloads hold companion `<id>-<hash>.lock` files, and both flows already remove their own directories when they finish. A recursive cleaner would need to participate in those lock protocols non-blockingly, which `starbase_utils` has no primitive for today. Happy to follow up on that separately (e.g. a try-lock variant in `starbase_utils` plus layout-aware temp cleanup) if there's interest.

Two small behavior corrections that fall out of the rewrite:

- The dot-entry check now compares bytes instead of going through `to_str()`, which returned `None` for non-UTF-8 names and caused such dot entries to be deleted rather than skipped.
- Stale file symlinks at the top level were previously deleted (`Path::is_file()` follows links); symlinks are now always left alone.

Why not `starbase_utils::fs::remove_dir_stale_contents()`: it returns aggregate counts rather than the per-file results the CLI reports, and has no dot-entry or pruning policy.

Tests cover nested stale cache cleanup with pruning, fresh nested files surviving, dot directories untouched, directory and file symlinks preserved, and temp remaining top-level only. Also verified manually against a 10 GB real-world `~/.proto/cache`.

### DIFF
--- a/crates/cli/src/commands/clean.rs
+++ b/crates/cli/src/commands/clean.rs
@@ -13,6 +13,7 @@ use starbase_console::ui::*;
 use starbase_console::utils::formats::format_bytes_binary;
 use starbase_styles::color;
 use starbase_utils::fs;
+use std::io;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, SystemTime};
 use tracing::{debug, instrument};
@@ -269,23 +270,26 @@ pub async fn clean_proto_tool(session: &ProtoSession, days: u64) -> miette::Resu
 }
 
 #[instrument]
-pub async fn clean_dir(dir: &Path, days: u64) -> miette::Result<Vec<StaleFile>> {
+pub fn clean_dir(dir: &Path, days: u64, recurse: bool) -> miette::Result<Vec<StaleFile>> {
     let duration = Duration::from_secs(86400 * days);
     let mut cleaned = vec![];
 
-    for file in fs::read_dir(dir)? {
-        let path = file.path();
+    for entry in fs::read_dir(dir)? {
+        let path = entry.path();
 
         // Don't delete dot files/folders
         if path
             .file_name()
-            .and_then(|name| name.to_str())
-            .is_some_and(|name| name.starts_with('.'))
+            .is_some_and(|name| name.as_encoded_bytes().starts_with(b"."))
         {
             continue;
         }
 
-        if path.is_file() {
+        let Ok(file_type) = entry.file_type() else {
+            continue;
+        };
+
+        if file_type.is_file() {
             let bytes = fs::remove_file_if_stale(&path, duration)?;
 
             if bytes > 0 {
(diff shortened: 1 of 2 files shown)


milesj:
Thanks!

PR #1049, opened 2026-07-14
new: Support better archive prefixing.

milesj:


milesj:
Going with a different approach.

PR #1055, opened 2026-07-20
fix: prevent reporter stdout corruption

### LINKED ISSUE #1054: Proto writes NDJSON records to command output in AI agent environments
## Version

0.58.2

## Commands

- `proto activate`
- `proto completions`
- `proto mcp`
- `proto bin`
- `proto exec`
- `proto run`
- `proto shell`

## Steps to reproduce

1. Set an AI agent environment variable.

   ```sh
   export CODEX_CI=1
   ```

2. Run this command.

   ```sh
   proto activate zsh
   ```

3. Evaluate the output in Zsh.

   ```zsh
   eval "$(proto activate zsh)"
   ```

The result is the same when `CLAUDECODE=1` is set.

## Actual behavior

`proto activate zsh` writes NDJSON records and Zsh code to standard output.

Zsh reads the NDJSON records as commands. Zsh reports an error similar to this one:

```text
command not found: {"type":"message",...}
```

Proto does not define `_proto_activate_hook`. Directory changes do not run the Proto activation hook.

With the same environment variable, other commands also write extra data to standard output:

- `proto activate --export` writes NDJSON records with shell export code.
- `proto activate nu --json` writes NDJSON instead of one JSON document.
- `proto completions` writes NDJSON records with completion code.
- `proto mcp` writes NDJSON records with JSON-RPC data.
- `proto bin` writes NDJSON records with a path.
- `proto exec`, `proto run`, and `proto shell` can write NDJSON records before child-process output.

An MCP tool installation can also write build user-interface output to the JSON-RPC stream.

## Expected behavior

Each command writes only its specified output format to standard output.

Shell code, completion code, JSON-RPC data, paths, and child-process output do not contain reporter records or user-interface output.

## Environment

- Operating system: macOS 26.5.2
- Architecture: arm64

liby:
## Summary

Proto automatically selects NDJSON in AI agent environments. Some commands reserve standard output for another data format. Reporter records can then corrupt shell code, completion code, JSON-RPC data, paths, and child-process output.

The CLI now assigns one owner to standard output for every command. The reporter writes to standard output only when it owns the stream. Explicit JSON output remains available for `proto activate` and `proto bin`.

Clap value sources define the priority of `--json`, `--reporter`, `PROTO_JSON`, and `PROTO_REPORTER`. Command-line values take priority over environment values. `--reporter` takes priority over `--json` when both values have the same source.

The Nushell hook requests `--reporter json`. Quiet installs do not render the build user interface in the MCP JSON-RPC stream.

Fixes #1054

## Test results

- `cargo fmt --all --check`
- `cargo clippy --package proto_cli --all-targets`
- `session`: 1 test passed.
- `activate_test`: 19 tests passed.
- `bin_test`: 7 tests passed.
- `completions_test`: 2 tests passed.
- `exec_test`: 10 tests passed.
- `general_test`: 6 tests passed.
- `mcp_test`: 1 test passed.

### DIFF
--- a/crates/cli/src/app.rs
+++ b/crates/cli/src/app.rs
@@ -216,9 +216,6 @@ impl App {
         self.json || self.reporter.is_some()
     }
 
-    /// Resolve the stdout owner for the current command. Exhaustive on
-    /// purpose: a new command must decide who owns its stdout, instead of
-    /// silently inheriting the reporter and its AI agent NDJSON fallback.
     pub fn stdout_owner(&self) -> StdoutOwner {
         match &self.command {
             Commands::Activate(args) => {
@@ -236,27 +233,7 @@ impl App {
                     StdoutOwner::McpStdio
                 }
             }
-            Commands::Alias(_)
-            | Commands::Bin(_)
-            | Commands::Clean(_)
-            | Commands::Debug { .. }
-            | Commands::Diagnose(_)
-            | Commands::Exec(_)
-            | Commands::Install(_)
-            | Commands::Migrate(_)
-            | Commands::Outdated(_)
-            | Commands::Pin(_)
-            | Commands::Plugin { .. }
-            | Commands::Regen(_)
-            | Commands::Run(_)
-            | Commands::Setup(_)
-            | Commands::Shell(_)
-            | Commands::Status(_)
-            | Commands::Unalias(_)
-            | Commands::Uninstall(_)
-            | Commands::Unpin(_)
-            | Commands::Upgrade(_)
-            | Commands::Versions(_) => StdoutOwner::Reporter,
+            _ => StdoutOwner::Reporter,
         }
     }
 
--- a/crates/cli/src/app.rs
+++ b/crates/cli/src/app.rs
@@ -6,7 +6,7 @@ use crate::commands::{
     plugin::{PluginAddArgs, PluginInfoArgs, PluginListArgs, PluginRemoveArgs, PluginSearchArgs},
 };
 use clap::builder::styling::{Color, Style, Styles};
-use clap::{Parser, Subcommand, ValueEnum};
+use clap::{ArgMatches, CommandFactory, FromArgMatches, Parser, Subcommand, ValueEnum};
 use proto_core::{ConfigMode, reporter::ReporterFormat};
 use starbase_styles::color::Color as ColorType;
 use std::{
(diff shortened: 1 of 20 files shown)


crates/cli/src/app.rs
```diff
@@ -199,8 +190,129 @@ pub struct App {
...
+pub enum StdoutOwner {
+    Reporter,
+    ShellCode,
+    CompletionCode,
+    McpStdio,
+}
+
 impl App {
-    pub fn setup_env_vars(&mut self) {
+    pub fn parse_with_reporter_precedence() -> Self {
```
milesj: Only part I don't really like is all of this. Ideally I don't want to eject out of what clap is doing. This is too custom.

liby: I removed the custom `ArgMatches` and `ValueSource` handling and restored the standard `CLI::parse()` path.

The remaining application-level logic only preserves the existing `--json` alias semantics: JSON-family reporters keep their exact format, the JSON alias upgrades a text reporter to JSON, and AI-agent detection remains the fallback.

I also added regression coverage for `PROTO_REPORTER=text` with `--json` and isolated the test defaults from developer shell settings.

crates/cli/src/session.rs
```diff
@@ -286,7 +291,10 @@ impl AppSession for ProtoSession {
     type Error = miette::Report;
 
     async fn startup(&mut self) -> AppResult<Self::Error> {
-        if self.cli.reporter == ReporterFormat::Ndjson && ai_env::is_ai_agent() {
+        if self.cli.stdout_owner() == StdoutOwner::Reporter
```
milesj: This should only show for AI agents as it is instructions.

crates/cli/src/app.rs
```diff
@@ -199,8 +198,78 @@ pub struct App {
...
+            }
+            Commands::Completions(_) => StdoutOwner::CompletionCode,
+            Commands::Mcp(args) => {
+                if args.info {
+                    StdoutOwner::Reporter
+                } else {
+                    StdoutOwner::McpStdio
+                }
+            }
+            Commands::Alias(_)
```
milesj: This can just be `_ => ...`? right?

milesj:
@liby Few comments and conflicts.

liby:
Updated both points and rebased onto the latest master. The notice now only appears in AI-agent environments, and `stdout_owner` uses `_ => StdoutOwner::Reporter` by default.

milesj:
Awesome thanks!

PR #1064, opened 2026-07-29
fix: re-check the install state after acquiring the install lock

### LINKED ISSUE #1063: Concurrent auto-installs unpack on top of a finished install, so other processes exec a half-written binary
### What version?

0.59.0

### Which command?

`proto run` (and the shims, whenever `auto-install` fires)

### What happened?

I work in tmux and open four or five panes in the same project, one right after another. Each shell enters the project, each one triggers an auto-install of the pinned Node version, and some of them fail like this:

```
/tmp/xfs-d7b44646/yarn: line 2: /home/talinux/.local/share/proto/tools/node/26.4.0/bin/node: Permission denied
```

The binary is there, it is just not finished. It sits at its final path with no execute bit and a size that keeps climbing if you stat it in a loop:

```
-rw-r--r--  138099400  bin/node    while it is being written
-rwxr-xr-x  147947264  bin/node    once the install completes
```

This is the `.prototools` of the project where the bug happens:

```toml
node = "~26"
yarn = "~4"

[settings]
    auto-clean        = true
    auto-install      = true
    telemetry         = false
    unstable-lockfile = true

[backends.node.env]
    file = ".env.development"

[backends.yarn.env]
    file = ".env.development"

[tools.node]
    bundled-npm = true
```

The version range is `~26`, so the exact version comes from `.protolock`, which pins `26.4.0`. That matters for reproducing: the machine had `26.5.0` installed and nothing at `26.4.0`, so entering the project made every shell want the same missing version at the same time.

The first pane installs Node and succeeds. The other four passed the `is_installed` check before that happened, so they block on the lock, and once they get it they download and unpack again, into the same directory the first install already finished. Nothing tells them the work is done.

`Installer::install` checks `is_installed` before taking the lock and never checks again after acquiring it ([`crates/core/src/flow/install.rs#L83`](https://github.com/moonrepo/proto/blob/master/crates/core/src/flow/install.rs#L83)):

```rust
if self.tool.is_installed(self.spec) && !options.force {
    return Ok(None);
}
// ...
let mut install_lock = fs::lock_directory(&self.temp_dir)?;
```

The unpack then writes straight into the live install directory, with no staging directory and no atomic move ([`install.rs#L505`](https://github.com/moonrepo/proto/blob/master/crates/core/src/flow/install.rs#L505)):

```rust
archive::unpack(&self.product_dir, &self.temp_dir, ...)
```

So `bin/node` is truncated back to a few kilobytes and grown again, and it stays non-executable until the extraction finishes.

Every other process on the machine sees a healthy install during that whole window. `is_installed` guards on a lock held over the install directory, but the install locks `temp_dir` instead, so that guard never fires for prebuilt installs ([`crates/core/src/tool.rs#L209`](https://github.com/moonrepo/proto/blob/master/crates/core/src/tool.rs#L209)):

```rust
let installed = spec.version.as_ref().is_some_and(|v| {
    !v.is_latest() && self.inventory.manifest.installed_versions.contains(v)
}) && dir.exists()
    && !fs::is_dir_locked(&dir);
```

The comment above the lock explains why it lives on `temp_dir`, so I assume moving it is not a simple change:

```rust
// Lock the temporary directory instead of the install directory,
// because the latter needs to be clean for "build from source",
// and the `.lock` file breaks that contract
```

Anything that execs the resolved path during one of those windows gets exit code 126. In my case it was the yarn launcher script, which had already resolved the path to node.

### Reproduction

This runs in a sandboxed `PROTO_HOME` and does not touch a real install. It pins the version directly instead of going through `.protolock`, since the lockfile only decides which version everyone races for.

```bash
#!/usr/bin/env bash
set -u

NODE_VERSION=26.4.0
SANDBOX="$(mktemp -d)"
export PROTO_HOME="$SANDBOX/proto"
WORK="$SANDBOX/project"
LOG="$SANDBOX/failures.log"
mkdir -p "$PROTO_HOME" "$WORK"

cat > "$WORK/.prototools" <<EOF
node = "$NODE_VERSION"

[settings]
auto-install = true
telemetry = false
EOF

BIN="$PROTO_HOME/tools/node/$NODE_VERSION/bin/node"
cd "$WORK" || exit 1

# Stand-in for anything that execs the resolved binary directly,
# such as the yarn launcher script or an editor that cached the path.
(
  while :; do
    if [ -e "$BIN" ]; then
      msg="$("$BIN" --version 2>&1)"
      code=$?
      if [ "$code" -ne 0 ]; then
        printf '%s exit=%s %s %s\n' \
          "$(date +%T.%3N)" "$code" "$(stat -c '%s %A' "$BIN")" "${msg##*: }" >> "$LOG"
      fi
    fi
    sleep 0.02
  done
) &
VICTIM=$!

# Five shells entering the project a moment apart.
for _ in 1 2 3 4 5; do
  proto run node -- --version > /dev/null 2>&1 &
  sleep 0.3
done
wait $(jobs -p | grep -v "^$VICTIM$") 2>/dev/null

sleep 1
kill "$VICTIM" 2>/dev/null

echo "failed execs: $(wc -l < "$LOG" 2>/dev/null || echo 0)"
head -3 "$LOG" 2>/dev/null
tail -1 "$LOG" 2>/dev/null
```

Output on my machine:

```
failed execs: 368
19:06:00.066 exit=126 182017 -rw-r--r-- Permission denied
19:06:00.090 exit=126 2141995 -rw-r--r-- Permission denied
19:06:00.114 exit=126 4783304 -rw-r--r-- Permission denied
19:06:12.980 exit=126 147881370 -rw-r--r-- Permission denied
```

Polling `stat` on the binary shows it reach the complete state five separate times, once per queued install. Four of those five rebuilds are pure waste, and each one leaves the binary unusable for roughly 1.7 seconds. That is about 13 seconds of broken Node out of a 20 second run. All five installs exit 0 and report success.

### Trace logs?

Two processes, started 0.3 seconds apart, trimmed to the lines that matter.

Process A:

```
23:06:32.510  Tool has not been installed
23:06:32.510  Locking directory  temp/node/MjYuNC4w
23:06:38.892  Attempting to unpack archive   install_dir: tools/node/26.4.0
23:06:41.026  Successfully installed tool    install_dir: tools/node/26.4.0
```

Process B:

```
23:06:32.806  Tool has not been installed
23:06:32.806  Locking directory  temp/node/MjYuNC4w
23:06:41.454  Attempting to unpack archive   install_dir: tools/node/26.4.0
23:06:44.113  Successfully installed tool    install_dir: tools/node/26.4.0
```

B blocked on the lock for 8.6 seconds, then unpacked into `tools/node/26.4.0` at `41.454`, which is 0.43 seconds after A had already declared that same directory successfully installed.

### What I would expect

Re-checking `is_installed` right after acquiring the lock would fix the common case and skip four redundant downloads. Unpacking into a staging directory and moving it into place would close the window for good, including the first install, but I understand the "build from source" constraint makes that less straightforward.

Related but not the same: #1057 was another concurrency bug between installs, fixed in 0.59.0. This one is about the install directory rather than the shim registry.

### Operating system?

Linux

### Architecture?

x64

ItzTas:
Fixes #1063.

Concurrent auto-installs of the same missing version all pass the `is_installed` check before any of them finishes. They then queue on the install lock and unpack again on top of a directory another process already installed into, which leaves the binary truncated and non-executable for a moment. Anything that execs the resolved path during that window fails with exit code 126.

This re-checks the install state after acquiring the lock. Reusing `Tool::is_installed` is not enough on its own, because the tool's manifest is loaded into memory before the wait, so `installed_versions` is still the copy from before the other process wrote to it. The check reads `manifest.json` from disk instead.

The re-check alone still left a race: the installing process released the lock inside `Installer::install`, but the manifest only reached disk later, inside `Linker::link`. A waiting process could acquire the lock in that gap, read the old manifest, and reinstall anyway. When I ran the repro from the issue with only the re-check applied, one of the four queued processes still got through. `Installer::install` now returns the lock guard along with the record, and `Manager::install` releases it only after the manifest is written to disk.

With both changes, the victim loop from the issue drops from 368 failed execs to zero once the initial extraction finishes, and only one of the five concurrent installers unpacks the archive.

### DIFF
--- a/crates/core/src/flow/install.rs
+++ b/crates/core/src/flow/install.rs
@@ -7,6 +7,7 @@ use crate::helpers::{is_archive_file, is_offline};
 use crate::lockfile::*;
 use crate::reporter::ProtoConsole;
 use crate::tool::Tool;
+use crate::tool_manifest::ToolManifest;
 use crate::tool_spec::ToolSpec;
 use crate::utils::log::LogWriter;
 use crate::utils::{archive, process};
@@ -79,7 +80,7 @@ impl<'tool> Installer<'tool> {
     pub async fn install(
         &self,
         options: InstallOptions,
-    ) -> Result<Option<LockRecord>, ProtoInstallError> {
+    ) -> Result<Option<(LockRecord, fs::DirLock)>, ProtoInstallError> {
         if self.tool.is_installed(self.spec) && !options.force {
             debug!(
                 tool = self.tool.context.as_str(),
@@ -97,6 +98,12 @@ impl<'tool> Installer<'tool> {
         // because the latter needs to be clean for "build from source",
         // and the `.lock` file breaks that contract
         let mut install_lock = fs::lock_directory(&self.temp_dir)?;
+
+        if !options.force && self.was_installed_while_locked() {
+            install_lock.unlock()?;
+
+            return Ok(None);
+        }
 
         // If this function is defined, it acts like an escape hatch and
         // takes precedence over all other install strategies
@@ -141,7 +148,7 @@ impl<'tool> Installer<'tool> {
                 // Verify against lockfile
                 Locker::new(self.tool).verify_locked_record(self.spec, &record)?;
 
-                return Ok(Some(record));
+                return Ok(Some((record, install_lock)));
             }
 
             if !output.skip_install {
@@ -182,11 +189,7 @@ impl<'tool> Installer<'tool> {
                     "Successfully installed tool",
                 );
 
-                install_lock.unlock()?;
-
-                let _ = fs::remove_dir_all(&self.temp_dir);
-
-                Ok(Some(record))
+                Ok(Some((record, install_lock)))
             }
 
             // Clean up if the install failed
@@ -205,6 +208,19 @@ impl<'tool> Installer<'tool> {
                 Err(error)
             }
         }
+    }
+
+    fn was_installed_while_locked(&self) -> bool {
+        let Some(version) = self.spec.version.as_ref() else {
+            return false;
+        };
+
+        if version.is_latest() || !self.product_dir.exists() {
+            return false;
+        }
+
+        ToolManifest::load_from(self.tool.get_inventory_dir())
+            .is_ok_and(|manifest| manifest.installed_versions.contains(version))
     }
 
     /// Build the tool from source using a set of requirements and instructions
--- a/crates/core/src/flow/manage.rs
+++ b/crates/core/src/flow/manage.rs
@@ -40,22 +40,24 @@ impl<'tool> Manager<'tool> {
             let cache_hit = self.tool.is_installed(spec) && !options.force;
             cache = cache_status(cache_hit);
 
-            let record = match Installer::new(self.tool, spec).install(options).await? {
-                // Update lock record with resolved spec information
-                Some(mut record) => {
-                    record.version = Some(version.clone());
-                    record.spec = Some(spec.req.clone());
-                    record
-                }
-                // Return an existing lock record if already installed
-                None => {
-                    self.post_install(spec, false).await?;
+            let (record, mut install_lock) =
+                match Installer::new(self.tool, spec).install(options).await? {
+                    // Update lock record with resolved spec information
+                    Some((mut record, install_lock)) => {
+                        record.version = Some(version.clone());
+                        record.spec = Some(spec.req.clone());
 
-                    return Ok(Locker::new(self.tool)
-                        .get_resolved_locked_record(spec)
-                        .cloned());
-                }
-            };
+                        (record, install_lock)
+                    }
+                    // Return an existing lock record if already installed
+                    None => {
+                        self.post_install(spec, false).await?;
+
+                        return Ok(Locker::new(self.tool)
+                            .get_resolved_locked_record(spec)
+                            .cloned());
+                    }
+                };
 
             // Add record to lockfile
             if spec.update_lockfile {
@@ -71,6 +73,9 @@ impl<'tool> Manager<'tool> {
                     ..Default::default()
                 },
             );
+            self.tool.inventory.manifest.save()?;
+
+            install_lock.unlock().map_err(ProtoInstallError::from)?;
 
             self.post_install(spec, true).await?;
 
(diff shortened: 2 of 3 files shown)


milesj:
@ItzTas So this _did_ work, and I know why it stopped working. So in the `is_installed` function, we check if the directory is locked: https://github.com/moonrepo/proto/blob/master/crates/core/src/tool.rs#L209

So if "locked", it's not installed. BUT I forgot I moved the lock from the product dir to the temp dir: https://github.com/moonrepo/proto/blob/master/crates/core/src/flow/install.rs#L99 So this lock check NEVER hits.

I think the correct fix is to check the correct dir, AND after the lock is acquired, check if it was installed via naother process.

ItzTas:
Pushed the changes. `is_installed` now checks the lock on the temp dir, the same dir `Installer::install` locks, so the lock check actually hits now. While one process is installing, every other process sees the version as not installed, and `proto run` waits on the lock instead of running a half-written binary. In the repro, everything going through proto stopped failing, even during the first extraction.

I tried to use `is_installed` for the check after the lock is acquired, and it can't work there. Two reasons:

1. The manifest was loaded into memory before the wait started. A version that another process installed while we waited is never in `installed_versions`, so the call always comes back false. The check needs to read `manifest.json` from disk, which is what `was_installed_while_locked` does.
2. With the lock check on the temp dir, `is_installed` is also false for the process that holds the lock. `is_dir_locked` tries a shared lock on the `.lock` file, and flock sees our own exclusive lock as contention, even inside the same process.

The stress tests caught one more problem. A process that skips the install still has its old manifest in memory, and `link_shims` saves that manifest when it creates shims. With several processes racing through `post_install`, one of them can overwrite `manifest.json` with an empty `installed_versions`, and the next `proto run` reinstalls everything. So `Manager::install` now reloads the manifest from disk before `post_install` in that branch.

In the stress runs, five concurrent auto-installs unpack exactly once and nothing that goes through proto fails at any point. A SIGKILL in the middle of an unpack also recovers: the flock probe ignores the stale `.lock` file, and the next install just redoes the work.

milesj:
I've been debugging this with fable and it turns out there are other deep related issues, so I fixed them all here: https://github.com/moonrepo/proto/pull/1071

Thanks for bringing it to my attention.

PR #1085, opened 2026-08-18
fix: Do not append trailing separators when converting virtual path prefixes

RemiKalbe:
When the path being converted through `convert_to_virtual_path`, `convert_to_real_path`, or `convert_to_real_native_path` is exactly one of the mapped prefixes, `strip_prefix` yields an empty relative path, and `Path::join("")` appends a trailing separator:

```rust
PathBuf::from("/Users/me/repo").join("") // => "/Users/me/repo/"
```

So converting a workspace root round-trips as `/workspace/` → `/Users/me/repo/`. Downstream in moon, that slashed path is exported as `PWD` for plugin-executed commands, and shells that validate `PWD` on startup (nushell) refuse to run, failing the command. Full analysis in moonrepo/moon#2676.

This went unnoticed because `PathBuf` equality compares components and ignores trailing separators, so the existing `converts_the_prefixes_themselves` test passes even with the bug present.

### Changes

- Added a prefix equality branch to all three convert functions that returns the opposite prefix directly, so `join` only runs for strictly longer paths. Since `Path` equality is component-based, this also covers inputs that already carry a trailing separator.
- Added tests that assert on the raw string form (unix and windows variants), including inputs that themselves carry a trailing separator. Verified the new test fails without the fix.
- Added a changelog entry.

### DIFF
--- a/crates/warpgate-api/src/path_utils.rs
+++ b/crates/warpgate-api/src/path_utils.rs
@@ -34,6 +34,17 @@ pub fn sort_paths_list(paths_list: &mut [(PathBuf, PathBuf)]) {
     paths_list.sort_by(|a, d| d.0.cmp(&a.0).then(d.1.cmp(&a.1)));
 }
 
+// When the path being converted is the prefix itself, the stripped
+// relative path is empty, and `Path::join("")` would append a trailing
+// separator, so return the base prefix untouched instead.
+fn join_rel_path(base: &Path, rel_path: &Path) -> PathBuf {
+    if rel_path.as_os_str().is_empty() {
+        base.to_owned()
+    } else {
+        base.join(rel_path)
+    }
+}
+
 /// Convert the provided real host path to a virtual guest path.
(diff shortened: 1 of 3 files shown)


crates/warpgate-api/src/path_utils.rs
```diff
@@ -47,7 +58,7 @@ pub fn convert_to_virtual_path(
         let virtual_path = if path.starts_with(guest_path) {
             path.to_owned()
         } else if let Ok(rel_path) = path.strip_prefix(host_path) {
-            guest_path.join(rel_path)
+            join_rel_path(guest_path, rel_path)
```
milesj: I think all you need is a `path == host_path` check. Then you avoid operating on the os string.

milesj (changes requested):

crates/warpgate-api/src/path_utils.rs
RemiKalbe: Done

Issue #1057, opened 2026-07-21
Shim registry (shims/registry.json) loses all existing entries when a concurrent proto install races the empty-JSON fallback


IgnisDa:
## Summary

`~/.proto/shims/registry.json` is a single file shared by **every** installed tool. When a `proto install` reads that file at the exact moment a concurrent `proto install` (for a *different* tool) has truncated it but not yet finished writing, the reader falls back to an empty registry (`{}`) instead of retrying or erroring — and then persists that near-empty registry back to disk, silently **deleting the shim metadata for every other previously-installed tool**.

The most visible symptom: secondary executables like `uvx`, `bunx`, and `npx` stop working with:

```
Unable to proceed, uvx is not a built-in plugin and has not been configured with [plugins] in a .prototools file.
```

even though `uv`/`bun`/`npm` themselves keep working fine (they're built-ins with a version pin; `uvx`/`bunx`/`npx` are secondary executables that only resolve via the `context`/`alt_exe` metadata in the shim registry).

## Environment

- proto `0.58.2`
- macOS (aarch64-apple-darwin), but the race is not macOS-specific — nothing in the code path is platform-gated except the mitigating lock (see below)

## Root cause

1. **[`crates/core/src/helpers.rs#L131-L137`](https://github.com/moonrepo/proto/blob/aaf01083521924cfbe2702c6ed705043f8177cbf/crates/core/src/helpers.rs#L131-L137)** — `read_json_file_with_lock` falls back to an empty object when it reads a zero-length file:

   ```rust
   // When multiple processes are ran in parallel, we may run into an issue where
   // the file has been truncated, so JSON parsing fails. It's a rare race condition,
   // and these file locks don't seem to catch it. If this happens, fallback to empty JSON.
   // https://github.com/moonrepo/proto/issues/85
   if content.is_empty() {
       content = "{}".into();
   }
   ```

[...]

PR #1096, opened 2026-08-21
fix: Merge concurrent `.prototools` writes instead of overwriting

### LINKED ISSUE #1095: Concurrent `.prototools` writes erase each other's entries
### Summary

Two processes that mutate the same `.prototools` can erase each other's entries. `ProtoConfig::update_document` reads the file under a shared lock, releases it, then re-acquires an exclusive lock to write, so both processes can read the same content and the last writer wins. It is the writer behind every config mutation: `install --pin`, `pin`, `unpin`, `alias`, `unalias`, `uninstall`, `outdated --update`, `plugin add` and `plugin remove`.

https://github.com/moonrepo/proto/blob/7ce999e3/crates/core/src/config.rs#L498-L519

### How it showed up

The E2E suite hit it on master: https://github.com/moonrepo/proto/actions/runs/32400669447/job/96527876173 (`18-install-moon` and `32-npm-backend` failed on `ubuntu-latest`, both green on macOS and Windows). The suite runs `10`–`18` in one parallel group, and each of those installs with `--pin local` into the same scratch `.prototools`.

From the uploaded logs, the write window is visible:

```
12-install-deno.log:  [TRACE 18:07:14.275] Acquired shared lock     file=".../.prototools"   <- reads
18-install-moon.log:  [TRACE 18:07:14.279] Writing file             file=".../.prototools"   <- moon lands
12-install-deno.log:  [TRACE 18:07:14.279] Writing file             file=".../.prototools"   <- deno overwrites
```

`deno` read before `moon` wrote, so `moon = "2.2.6"` was gone from the config a few milliseconds after being pinned. Both failing tests are the two that then invoke the `moon` shim.

The interesting part is how the symptom presented, because it is what a user would report:

- `proto bin moon` **succeeded** and printed `.../moon/2.2.6/moon`. `bin` tolerates a failed detection (`ToolRecord::detect_version_and_source` swallows the error) and falls back to `*`, which resolves to the newest installed version.
- The `moon` shim **failed**. `proto run` does not tolerate it: `Detector::detect` returns `FailedVersionDetect`, `run_global_tool` finds no global `moon` on a CI runner, and the detect error is returned.

So the install logged success, `proto bin` agreed the tool was there, and only `proto run` disagreed — with the config as the only evidence, and nothing in it pointing at a lost write.

### Repro

```bash
cd "$(mktemp -d)"
for tool in node bun deno go; do proto pin "$tool" latest & done; wait
cat .prototools     # typically 1 of the 4 entries
```

### Impact

Any script that installs tools in parallel and pins them — a CI job fanning out `proto install <tool> --pin`, or a Makefile/justfile doing the same — silently drops pins. The failure surfaces later and elsewhere, as `proto run` failing to detect a version for a tool that just installed successfully.

### Fix

Hold a single exclusive lock across read → parse → mutate → write, which is what `ShimRegistry::save` already does for `registry.json` after the same class of bug (#1057). PR: <PR_A>

`ToolManifest::load`/`save` (`crates/core/src/tool_manifest.rs`) has the same read-then-write shape. It is per-tool rather than shared by all of them, so the window is much narrower — noting it here rather than changing it.

### Why this took a while to diagnose

The E2E logs for both failures end on `Verifying shim version...` with no exit code, no output and no assertion message — 290 KB of trace logging and the actual error nowhere in it. `e2e/lib/utils.sh` does:

```bash
ver=$("$shim" "$version_arg" 2>&1)   # set -e aborts the test here
shim_rc=$?                           # never runs
echo "  exit=$shim_rc"               # never runs
echo "  output=$ver"                 # never runs, the error dies with $ver
```

Under `set -e` a failing command substitution aborts immediately, so every `if [[ $rc -ne 0 ]]` below it is dead code. A second PR fixes that, adds a stack-trace `ERR` trap, dumps the shared config and shim registry on failure, and adds `E2E_SERIAL=1` for exactly this kind of suspected race: <PR_B>

KeyZer:
Closes #1095.

`ProtoConfig::update_document` read the config with a shared lock, released it, then re-acquired an exclusive lock to write. Two processes could therefore both read the same content, and whichever wrote last erased every entry the other had added. It is the writer behind every config mutation: `install --pin`, `pin`, `unpin`, `alias`, `unalias`, `uninstall`, `outdated --update`, `plugin add` and `plugin remove`.

The read, parse, mutation and write now happen while holding a single exclusive lock, mirroring `ShimRegistry::save` after the same class of bug (#1057). Readers take a shared lock, so they block on ours rather than observing the truncated file. `FileLock` needs a handle it can read as well as write, which `fs::create_file_if_missing` does not give us, hence `open_config_for_update`.

### Test

`crates/cli/tests/pin_concurrent_test.rs` pins 4 tools from 4 processes while the test itself holds the config lock, so all of them are queued on the same read-modify-write before any can finish. It also warms the plugin cache first — without that, the processes spend seconds compiling WASM and reach the lock too far apart to contend.

Without the fix, 3 of the 4 pins vanish:

```
missing protostar in config:
protojoin = "5.0.0"
moonstone = "4.0.0"
```

### Notes

- `ProtoConfig::update` has the same shape, but every call site today is a test. Left alone; happy to change it too if you'd rather they share one path.
- Tested on macOS and Linux locally. The Windows leg is the one I can't try here: `FileLock::new` carries an `ERROR_ACCESS_DENIED` retry loop around file creation, which I did not replicate in `open_config_for_update` since byte-range locks don't deny `open` and `.prototools` is never deleted underneath us. Shout if CI disagrees.
- Surfaced by the E2E failure on master (run 32400669447), where 9 parallel `install --pin local` runs into one `.prototools` dropped moon's pin ~4ms after it was written. The issue has the traces.


KeyZer:
Worth noting for review: #1097 is the same suite on the same commit base with the harness changes but without this fix, and its `Suite (ubuntu-latest)` leg failed on the race — `17-install-uv` pinned `uv = "0.11.33"`, a peer install overwrote the config, and the `uv` shim then died with `proto::detect::failed`. Both `Suite (ubuntu-latest)` and `Suite (macos-latest)` are green here.

Evidence, including the post-mortem showing the config with 8 tools and no `uv`: https://github.com/moonrepo/proto/issues/1095#issuecomment-5374387938

KeyZer:
`Test (windows-latest)` was red on the first push and I've fixed it in 602a729, but the reason is worth your attention because it says something about the design, not just the test.

The test failure itself was scaffolding. It holds an exclusive lock on `.prototools` so the spawned processes queue on the read-modify-write; on Windows byte-range locks are mandatory rather than advisory, so that lock made the lock-free reads in `ProtoFileManager::load` (`ProtoConfig::load(path, false)`, file_manager.rs:74,84) fail with a lock violation instead of queueing. Every process died during startup, before reaching the pin. The race isn't platform specific and the scaffolding can't work there, so the file is now gated like `activate_windows_test.rs`.

The part worth deciding on: that same interaction exists outside the test. While any writer holds the exclusive lock, a concurrent proto command reading that config on Windows fails rather than waits. That is true on master too — `write_file_with_lock` holds an exclusive lock across truncate + write — but this PR widens the window from a write to a read + parse + write, so it makes an existing hazard slightly more likely.

Three ways to go, and I'd rather you pick than guess:

1. **As it stands.** Simplest, fixes the lost update, leaves the Windows read-vs-lock hazard where master already had it.
2. **Sidecar lock + atomic rename**, exactly what `ShimRegistry::save` does for `registry.json`. The config file itself is never locked, so lock-free readers are never affected, and readers can't see a partial write either. The open question is where the lock file goes: next to `.prototools` litters the user's repo, so it would want to live in the store, which means `update_document` needs the store path passed in — a signature change across 8 call sites.
3. **Optimistic retry**: read under a short lock, parse and mutate unlocked, then take the lock, re-read, and only write if the content is unchanged, retrying otherwise. Keeps the locked window as short as master's, but `op` would have to become `FnMut` since it can run more than once, and "your closure may be called twice" is a contract I wouldn't add without your say-so.

Happy to implement 2 or 3 in this PR if you prefer either.

milesj:
I've reworked the locking impl in v0.61, so we'll see how that goes.

PR #1097, opened 2026-08-21
test(e2e): Never let a failing command hide its output

KeyZer:
Refs #1095, independent of #1096 (either can land first).

The E2E logs for both failures in run 32400669447 end on `Verifying shim version...` — no exit code, no output, no assertion message, 290 KB of trace logging and the actual error nowhere in it. The cause is in `lib/utils.sh`:

```bash
ver=$("$shim" "$version_arg" 2>&1)   # set -e aborts the test here
shim_rc=$?                           # never runs
echo "  exit=$shim_rc"               # never runs
echo "  output=$ver"                 # never runs, the error dies with $ver
```

Under `set -e` a failing command substitution aborts immediately, so every `if [[ $rc -ne 0 ]]` below it was dead code, and the same pattern was in `test_bin`, `install_tool` and `install_backend`.

### Changes

- **`run_probe` / `echo_probe`** (`lib/assert.sh`) — run a command, record `RUN_RC` / `RUN_OUT` / `RUN_ERR` keeping the two streams apart, never abort the caller. `test_bin` and `test_shim` now log the exit code, the output and the stderr before asserting on any of it. Keeping stderr separate also means `proto bin`'s path can't be polluted by whatever it logs.
- **`ERR` trap with a stack trace** (`lib/assert.sh`) — any unhandled failure now prints the command, `file:line` and the call stack. Failures inside a command substitution are skipped (`BASH_SUBSHELL`), as they are either handled by the caller or reported again by the assignment around them, so tests like `41-bin-shim` that deliberately tolerate failures stay quiet.

  ```
  ASSERT FAIL: expected substring '999' in: moon 2.2.6
  COMMAND FAILED (exit=1) at e2e/lib/assert.sh:49: return 1
    called from assert_contains (e2e/lib/utils.sh:112)
    called from test_shim (e2e/lib/utils.sh:150)
    called from install_tool (e2e/tests/18-install-moon.sh:6)
  ```

- **Post-mortem on failure** (`run.sh`) — writes `.logs/<name>.postmortem.txt` with the shared `.prototools`, the shim registry, the shims dir and `proto status`, captured while the state still exists. None of that is visible in a test's own log, and it is where cross-test failures actually live. The failure report also leads with the `ASSERT FAIL` / `COMMAND FAILED` lines instead of burying them under the tail.
- **GitHub Actions annotations + step summary** (`run.sh`) — a red E2E job showed nothing beyond `Process completed with exit code 1`, not even which test failed. Failures now emit `::error` annotations carrying the assertion message, and the step summary lists every failed and skipped test with its cause. Both are guarded on `GITHUB_ACTIONS` / `GITHUB_STEP_SUMMARY`, so a local run is unchanged.
- **`E2E_SERIAL=1`** (`run.sh`) — never run a group in parallel. The first thing to reach for when a failure smells like tests interfering with their peers.
- **`E2E_TAIL`, `E2E_KEEP_SCRATCH`** (`run.sh`) — inline tail size, and keeping the shared work dir so it can be poked at afterwards.
- **`PROTO_HOME` defaults to `e2e/.proto-home`** (`lib/env.sh`) — it pointed at `$HOME/.proto`, so `just test-e2e` installed 20+ tools into the developer's own store, then `90-uninstall` removed their `uv` and `99-clean` pruned it. `e2e/.gitignore` already lists `.proto-home/` and the README already documented that location, so this is the drift being closed rather than a new choice. `E2E_USE_REAL_HOME=1` opts back in.
- **README** — `PROTO_BIN_DIR` doesn't exist in any script, the group names were stale (`install-base` / `install-deps` vs `tools` / `tools-secondary` / `backends`), and the store is not "wiped at the start of each run". Documents the knobs and how to debug a failure.

### Also found while testing this

Running the suite from inside an AI agent (Claude Code, Codex, Cursor) failed on `assert_executable`, because proto detects the agent and switches its output to NDJSON — so `proto bin node` returned a JSON envelope instead of a path. `lib/env.sh` now blanks the same env vars `crates/core/src/test_utils.rs` does and pins `PROTO_REPORTER=text`, so the suite behaves the same whoever runs it.

### Deliberately not changing

The shared scratch cwd. Nine parallel installs pinning into one `.prototools` is what caught a real lost-update bug in proto; isolating each test would have hidden it.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -21,9 +21,39 @@ bash ./e2e/run.sh install
 bash ./e2e/run.sh node
 ```
 
-Per-test stdout/stderr is captured under `e2e/.logs/<name>.log`. The shared `PROTO_HOME` lives at `e2e/.proto-home` and is wiped at the start of each run.
+Per-test stdout/stderr is captured under `e2e/.logs/<name>.log`. The shared `PROTO_HOME` lives at `e2e/.proto-home`, so a run never touches your own `~/.proto` — the suite installs 20+ tools into the store, uninstalls one, and runs `proto clean` over it.
 
-Override the binary location with `PROTO_BIN_DIR=/some/dir bash ./e2e/run.sh` if you want to test a binary other than `target/release/proto`.
+State is kept between runs (`bootstrap.sh` puts the binary under test into the store, so the harness can't wipe it). To start clean:
+
+```bash
+rm -rf e2e/.proto-home && ./e2e/bootstrap.sh
+```
+
+## Env knobs
+
+| Variable              | Effect                                                                                  |
+| --------------------- | --------------------------------------------------------------------------------------- |
+| `E2E_SERIAL=1`        | Never run a `# group:` in parallel. First thing to try when a failure smells like a race |
+| `E2E_TAIL=200`        | How many lines of a failing test's log to print inline (default 40)                      |
+| `E2E_KEEP_SCRATCH=1`  | Keep the shared work dir instead of deleting it on exit, and print its path              |
+| `E2E_USE_REAL_HOME=1` | Use `~/.proto` as the store instead of `e2e/.proto-home`                                 |
+
+## Debugging a failure
+
+A failing test prints, in order: the `ASSERT FAIL` / `COMMAND FAILED` lines from its log, the tail of the log, and the path to a post-mortem file. The post-mortem (`e2e/.logs/<name>.postmortem.txt`) captures the state the test left behind — the shared `.prototools`, the shim registry, the shims directory, and `proto status` — none of which is visible in the test's own log. CI uploads both as the `e2e-logs-<os>` artifact.
+
+Helpers never swallow a command's output. Use `run_probe` for anything whose exit code you want to assert on:
+
+```bash
+run_probe proto bin node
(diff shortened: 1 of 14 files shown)


KeyZer:
The `Suite (ubuntu-latest)` leg here is red, and it is the bug in #1095 rather than anything in this PR: this branch has the harness changes but not the fix, so the race is still live. It hit `17-install-uv` this time.

It is also the clearest demonstration of why this PR exists. Everything needed to diagnose it is in the run without opening a single log:

- The run page annotation reads `E2E (linux) 17-install-uv: ASSERT FAIL: .../shims/uv --version exited 1`.
- The log records the error the old harness discarded — `proto::detect::failed`, "Failed to detect an applicable version to run uv with".
- `17-install-uv.postmortem.txt` shows the shared `.prototools` holding 8 tools and no `uv`, moments after `uv = "0.11.33"` was pinned.

The same two legs are green on #1096, which carries the fix. So this one should go green once #1096 lands — happy to rebase on top of it if you'd rather see that before merging, or it can just wait its turn.

milesj:
@KeyZer Can you rebase master and try again.

PR #1101, opened 2026-08-24
internal: Remove third-party plugins with dead locators

milesj:
## What changed

Audited every `locator` in `registry/data/third-party.json` and removed the entries that can no longer be installed. The dataset goes from 135 → 100 plugins.

**Removed 35 entries:**

- **30 entries from `Phault/proto-toml-plugins`** — the entire repository no longer exists (`github.com/Phault/proto-toml-plugins` is a 404), so every locator under it 404s: actionlint, bazel, biome, caddy, cmake, cosign, dagger, dprint, fly, gitleaks, gum, hurl, hyperfine, infisical, jira, just, mage, mkcert, ninja, octopus, oxlint, rattler-build, ruff, shellcheck, shfmt, task, traefik, trufflehog, uv, wizer.
- **2 `hugo` entries** (standard + extended) — the `z0rrn/proto-plugins` repository no longer exists.
- **chester** — the repository is alive but `proto.toml` was deleted from it (no plugin file anywhere in the tree).
- **sdkman** (`github://ageha734/proto-sdkman`) and **terraform** (`github://ageha734/proto-terraform`) — not 404s. Both repositories exist with source, but have zero releases/tags, so the `github://` locator has no wasm artifact to resolve and installation can never succeed. These would become valid again if those authors publish a release.

**One fix instead of a removal:**

- **strawberry-perl** pointed at a `github.com/.../blob/...` URL, which serves an HTML page rather than TOML — broken as a locator. The plugin itself is still maintained, so the entry was repointed at the equivalent `raw.githubusercontent.com` URL instead of being deleted.

## How it was verified

- All 114 `https://` locators were fetched (following redirects), checking both the status code and the response body — that body check is what caught `strawberry-perl` returning HTML with a 200.
- All 21 `github://` locators were resolved through the GitHub API, confirming the repository exists and has at least one release carrying a `.wasm` asset.
- `cargo run -p proto_codegen` runs clean and produces no reformatting, so the file remains valid against the schema.

## Notes for reviewers

`gitleaks`, `just`, `ruff`, `terraform`, `yq` and `jq` had duplicate entries, so those tools still have a working alternative in the registry. However, **29 ids now have no entry at all** in either registry data file, including some popular ones: `uv`, `task`, `biome`, `shellcheck`, `shfmt`, `dprint`, `oxlint`, `bazel`, `actionlint`, `cmake`, `caddy`, `cosign`, `dagger`, `fly`, `gum`, `hugo`, `hurl`, `hyperfine`, `infisical`, `jira`, `mage`, `mkcert`, `ninja`, `octopus`, `rattler-build`, `traefik`, `trufflehog`, `wizer`, `chester`, `sdkman`. This is mostly the fallout of a single contributor's repository disappearing — sourcing replacement plugins for those tools is worth a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/registry/data/third-party.json
+++ b/registry/data/third-party.json
@@ -17,20 +17,6 @@
       ]
     },
     {
-      "id": "actionlint",
-      "locator": "https://raw.githubusercontent.com/Phault/proto-toml-plugins/main/actionlint/plugin.toml",
-      "format": "toml",
-      "name": "actionlint",
-      "description": "Static checker for GitHub Actions workflow files.",
-      "author": "Phault",
-      "homepageUrl": "https://github.com/rhysd/actionlint",
-      "repositoryUrl": "https://github.com/Phault/proto-toml-plugins",
-      "devicon": "githubactions",
-      "bins": [
-        "actionlint"
-      ]
-    },
-    {
       "id": "air",
       "locator": "https://raw.githubusercontent.com/jamesukiyo/proto-plugins/master/air.toml",
       "format": "toml",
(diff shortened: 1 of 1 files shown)


milesj:
@Phault Do you happen to still have all the previous TOML files so that we can host them ourselves?

Phault:
Hey @milesj 

So sorry about this, the repo had been archived for some time and I had no idea if it was actually being used.

Yes, I still have the repo with full history. [Here's a git bundle (compressed as GitHub funnily enough does not allow .bundle).](https://github.com/user-attachments/files/31383097/proto-toml-plugins-main.bundle.zip)

milesj:
@Phault Thanks so much!

